PR #2, opened 2020-01-19
Consume envelope self, removing need for mutable Options

gegy1000:


Restioson:
I don't know if this fix would work anymore :l

Restioson:
This would need to be rebased to master for merge

Restioson:
Fixed in 619df4e

PR #3, opened 2020-01-20
Async Responder

Restioson:
- Adds `AsyncHandler` trait for handling messages asynchronously
- Adds `do_send_async` and `send_async` for sending messages that should be handled asynchronous
- Added some documentation (unfinished)
- Unresolved question (issue will be open): should we drop the synchronous `Handler` trait altogether to streamline the code and interface? This would mean that synchronous responders would be ~5% slower (according to my benchmarks).

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -1,4 +1,4 @@
-use crate::envelope::{Envelope, NonReturningEnvelope, SyncReturningEnvelope};
+use crate::envelope::{Envelope, NonReturningEnvelope, SyncReturningEnvelope, AsyncReturningEnvelope};
 use crate::{Actor, Handler, Message, SyncResponder};
 use futures::channel::mpsc::UnboundedSender;
 use futures::future::Either;
@@ -34,6 +34,27 @@ impl<A: Actor> Address<A> {
     {
         let t = SyncReturningEnvelope::new(message);
         let envelope: SyncReturningEnvelope<A, M> = t.0;
+        let rx = t.1;
+
+        let res = self
+            .sender
+            .unbounded_send(Box::new(envelope))
+            .map_err(|_| Disconnected);
+
+        match res {
+            Ok(()) => Either::Left(rx.map_err(|_| Disconnected)),
+            Err(e) => Either::Right(futures::future::err(e)),
+        }
+    }
+
+    pub fn send_async<M>(&self, message: M) -> impl Future<Output = Result<M::Result, Disconnected>>
+        where
+            M: Message,
+            A: Handler<M>,
+            A::Responder: Future<Output = M::Result> + Send,
+    {
+        let t = AsyncReturningEnvelope::new(message);
+        let envelope: AsyncReturningEnvelope<A, M> = t.0;
         let rx = t.1;
 
         let res = self
--- a/src/address.rs
+++ b/src/address.rs
@@ -1,4 +1,4 @@
-use crate::envelope::{Envelope, NonReturningEnvelope, SyncReturningEnvelope, AsyncReturningEnvelope};
+use crate::envelope::{Envelope, SyncNonReturningEnvelope, SyncReturningEnvelope, AsyncReturningEnvelope, AsyncNonReturningEnvelope};
 use crate::{Actor, Handler, Message, AsyncHandler};
 use futures::channel::mpsc::UnboundedSender;
 use futures::future::Either;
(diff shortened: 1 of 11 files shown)


src/address.rs
```diff
@@ -0,0 +1,87 @@
...
+        M: Message,
+        A: AsyncHandler<M>,
+    {
+        let envelope = AsyncNonReturningEnvelope::new(message);
+        self.sender
+            .unbounded_send(Box::new(envelope))
+            .map_err(|_| Disconnected)
+    }
+
+    pub fn send<M>(&self, message: M) -> impl Future<Output = Result<M::Result, Disconnected>>
```
gegy1000: There’s a bit of duplication here between `send` and `send_async`, is there any way to reduce that?

src/context.rs
```diff
@@ -0,0 +1,43 @@
+use crate::Actor;
+use std::marker::PhantomData;
+
+/// `Context` is used to signal things to the [`ActorManager`](struct.ActorManager.html)'s
+/// management loop. It can be used to stop the actor ([`Context::stop`](struct.Context.html#method.stop)) or
+/// queue a message to be processed by the actor immediately after it has finished processing the
+/// current message ([`Context::notify`](struct.Context.html#method.notify)).
+pub struct Context<A: Actor + ?Sized> {
+    pub(crate) running: bool,
+    phantom: PhantomData<A>, // TODO(weak_address)
```
gegy1000: For the TODO and commented out notifications, do you want to handle that in this PR?

src/envelope.rs
```diff
@@ -0,0 +1,160 @@
...
+use futures::{future, Future, FutureExt};
+use std::marker::PhantomData;
+use std::pin::Pin;
+
+type Fut<'a> = Pin<Box<dyn Future<Output = ()> + Send + 'a>>;
+
+pub(crate) trait Envelope: Send {
+    type Actor: Actor + ?Sized;
+
+    // We could return some enum like:
```
gegy1000: Is this comment still relevant?

src/envelope.rs
```diff
@@ -0,0 +1,160 @@
...
+    // do with branch [mis]prediction or compiler optimisation
+    fn handle<'a>(
+        &'a mut self,
+        act: &'a mut Self::Actor,
+        ctx: &'a mut Context<Self::Actor>,
+    ) -> Fut<'a>;
+}
+
+pub(crate) struct SyncReturningEnvelope<A: Actor + ?Sized, M: Message> {
+    message: Option<M>, // Options so that we can opt.take()
```
gegy1000: Shouldn't need `Option` here and a few other places with #2 -- do we want to merge that after?

src/envelope.rs
```diff
@@ -0,0 +1,160 @@
+use crate::{Actor, AsyncHandler, Context, Handler, Message};
+use futures::channel::oneshot::{self, Receiver, Sender};
+use futures::{future, Future, FutureExt};
+use std::marker::PhantomData;
+use std::pin::Pin;
+
+type Fut<'a> = Pin<Box<dyn Future<Output = ()> + Send + 'a>>;
+
+pub(crate) trait Envelope: Send {
```
gegy1000: Envelopes always return a future, does there need to be distinction between sync and async envelopes?

src/manager.rs
```diff
@@ -0,0 +1,74 @@
...
+            receiver,
+            actor,
+            ctx,
+        };
+        let addr = Address { sender };
+
+        (addr, mgr)
+    }
+
+    //    /// Handle notifications (messages to self)
```
gegy1000: These TODOs here too?

src/envelope.rs
Restioson: Yes.

src/envelope.rs
Restioson: Yes, it's about the `impl Envelope` thing here. If you want you can try use specialisation to make one envelope type (that can come in another PR). But I tried it and it's painful.

src/envelope.rs
Restioson: I don't know if #2 will work still so we will merge it after, yes

src/manager.rs
Restioson: The bottom TODO applies to this. Didn't want to duplicate it because when I implement notifications again I will know to look here.

src/context.rs
Restioson: No. Weak_address will be the next branch/pr name

src/address.rs
Restioson: I don't think so because of the bounds `Handler<M>` and `AsyncHandler<M>` having no common types, at least without macros. Not sure.

src/context.rs
gegy1000: What about notifications?

Restioson: Same thing ;P (weak address allows for notifications. it will be in the same pr)

PR #9, opened 2020-01-23
Weak addresses and various notifications

Restioson:
- Weak addresses that don't prevent the actor from being dropped
- `notify_interval` to notify every x duration
- `notify_after` to notify after x duration
- attach streams to addresses

Todo:
- [x] Check that weak address send.await works

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -4,9 +4,9 @@ use crate::{Actor, Handler, Message, MessageChannel, WeakMessageChannel};
 use futures::channel::mpsc::UnboundedSender;
 use futures::channel::oneshot::Receiver;
 use futures::task::{Context, Poll};
-use futures::{Future, FutureExt, Sink};
+use futures::{Future, Sink};
 #[cfg(any(doc, feature = "with-tokio-0_2", feature = "with-async_std-1"))]
-use futures::{Stream, StreamExt};
+use futures::{Stream, StreamExt, FutureExt};
 use std::pin::Pin;
 use std::sync::{Arc, Weak};
 
@@ -121,15 +121,17 @@ impl<A: Actor + Send> Address<A> {
     }
 
     pub fn channel<M: Message>(&self) -> MessageChannel<M>
-        where A: Handler<M>
+    where
+        A: Handler<M>,
     {
         MessageChannel {
-            address: Box::new(self.clone())
+            address: Box::new(self.clone()),
         }
     }
 
     pub fn into_channel<M: Message>(self) -> MessageChannel<M>
-        where A: Handler<M>
+    where
+        A: Handler<M>,
     {
         self.channel()
     }
@@ -238,15 +240,17 @@ pub struct WeakAddress<A: Actor> {
 
 impl<A: Actor + Send> WeakAddress<A> {
     pub fn channel<M: Message>(&self) -> WeakMessageChannel<M>
-        where A: Handler<M>
+    where
+        A: Handler<M>,
     {
         WeakMessageChannel {
-            address: Box::new(self.clone())
+            address: Box::new(self.clone()),
         }
     }
 
     pub fn into_channel<M: Message>(self) -> WeakMessageChannel<M>
-        where A: Handler<M>
+    where
+        A: Handler<M>,
     {
         self.channel()
     }
--- a/src/address.rs
+++ b/src/address.rs
@@ -88,6 +88,8 @@ impl<A: Actor> Address<A> {
 }
 
 impl<A: Actor> AddressExt<A> for Address<A> {
+    // To read more about what an envelope is and why we use them, look under `envelope.rs`
+
     fn do_send<M>(&self, message: M) -> Result<(), Disconnected>
     where
         M: Message,
@@ -227,6 +229,8 @@ impl<A: Actor> Clone for WeakAddress<A> {
     }
 }
 
-/// The actor is no longer running and disconnected
+/// The actor is no longer running and disconnected from the sending address. For why this could
+/// occur, see the [`Actor::stopping`](trait.Actor.html#method.stopping) and
+/// [`Actor::stopped`](trait.Actor.html#method.stopped) methods.
 #[derive(Clone, Eq, PartialEq, Debug)]
 pub struct Disconnected;
--- a/src/address.rs
+++ b/src/address.rs
@@ -66,9 +66,11 @@ pub trait AddressExt<A: Actor> {
 }
 
 /// An `Address` is a reference to an actor through which [`Message`s](trait.Message.html) can be
-/// sent. It can be cloned, and when all `Address`es are dropped, the actor will be stopped. It is
-/// created by calling the [`Actor::start`](trait.Actor.html#method.start) or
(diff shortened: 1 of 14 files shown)


src/address.rs
```diff
@@ -1,91 +1,338 @@
...
+    /// and does not wait for a response. If this returns `Err(Disconnected)`, then the actor is stopped
+    /// and not accepting messages. If this returns `Ok(())`, the will be delivered, but may
+    /// not be handled in the event that the actor stops itself (by calling [`Context::stop`](struct.Context.html#method.stop))
+    /// before it was handled.
+    fn do_send_async<M>(&self, message: M) -> Result<(), Disconnected>
+    where
+        M: Message,
+        A: AsyncHandler<M> + Send;
+
+    /// Sends a [`Message`](trait.Message.html) that will be handled asynchronously to the actor,
```
gegy1000: I think copy paste error here? Probably meant to be synchronously :P

src/address.rs
```diff
@@ -1,91 +1,338 @@
...
+        for<'a> A::Responder<'a>: Future<Output = M::Result> + Send;
+
+    /// Attaches a stream to this actor such that all messages produced by it are forwarded to the
+    /// actor and handled synchronously. This could, for instance, be used to forward messages from a
+    /// socket to the actor (after the messages have been appropriately `map`ped). This is a
+    /// convenience method over a manual async stream forwarding loop.
+    #[doc(cfg(feature = "with-tokio-0_2"))]
+    #[doc(cfg(feature = "with-async_std-1"))]
+    #[cfg(any(doc, feature = "with-tokio-0_2", feature = "with-async_std-1"))]
+    fn attach_stream<S, M>(self, mut stream: S)
```
gegy1000: You could possibly impl `Sink` for `T: AddrExt` which would allow `.forward` notation. Also then don't have to interact with tokio/async_std directly

gegy1000: If that's not possible, what about making it an async fn that has to be executed by the caller? Removes the need for duplication and gives user more control :)

src/address.rs
```diff
@@ -1,91 +1,338 @@
...
+/// methods.
+pub struct Address<A: Actor> {
+    pub(crate) sender: UnboundedSender<ManagerMessage<A>>,
+    pub(crate) ref_counter: Arc<()>,
+}
+
+impl<A: Actor> Address<A> {
+    /// Create a weak address to the actor. Unlike with the strong variety of address (this kind),
+    /// an actor will not be prevented from being dropped if only weak addresses exist.
+    pub fn weak(&self) -> WeakAddress<A> {
```
gegy1000: What about using same naming scheme as `Rc` for weak addresses? That being `upgrade` and `downgrade`?

src/address.rs
```diff
@@ -1,91 +1,338 @@
...
+impl<A: Actor> AddressExt<A> for WeakAddress<A> {
+    fn do_send<M>(&self, message: M) -> Result<(), Disconnected>
+    where
+        M: Message,
+        A: Handler<M> + Send,
+    {
+        // Check that there are external strong addresses. If there are none, the actor is
+        // disconnected and our message would interrupt its dropping. strong_count() == 2 because
+        // Context and manager both hold a strong arc to the refcount
+        if self.ref_counter.strong_count() > 2 {
```
gegy1000: I don't know if it can be avoided but it is a bit confusing to have 2 here be the minimum count. Maybe even just an internal function to get the actual number of strong counts (`ref_counter.strong_count() - 2`) would be nice

src/context.rs
```diff
@@ -1,26 +1,226 @@
...
 /// management loop. Currently, it can be used to stop the actor ([`Context::stop`](struct.Context.html#method.stop)).
 pub struct Context<A: Actor> {
+    /// Whether the actor is running. It is changed by the `stop` method as a flag to the `ActorManager`
+    /// for it to call the `stopping` method on the actor
     pub(crate) running: bool,
-    phantom: PhantomData<A>, // TODO(weak_address)
+    /// The address kept by the context to allow for the `Context::address` method to work.
+    address: Address<A>,
+    /// Notifications that must be stored for immediate processing.
+    pub(crate) immediate_notifications: Vec<Box<dyn Envelope<Actor = A>>>,
```
gegy1000: I hate to say it but would a `LinkedList` be a good option here? xP

src/context.rs
```diff
@@ -1,26 +1,226 @@
...
+    /// Stop the actor as soon as it has finished processing current message. This will mean that the
+    /// [`Actor::stopping`](trait.Actor.html#method.stopping) method will be called.
+    /// If that returns [`KeepRunning::No`](enum.KeepRunning.html#variant.No), any subsequent attempts
+    /// to send messages to this actor will return the [`Disconnected`](struct.Disconnected.html) error.
     pub fn stop(&mut self) {
         self.running = false;
     }
+
+    /// Get an address to the current actor if the actor is still running.
+    pub fn address(&self) -> Option<Address<A>> {
```
gegy1000: Can it ever be not running? Or is it possible to guarantee that the context always references a running actor? (I don't know if this is viable just wanna ask questions if they're maybe useful xP)

src/address.rs
Restioson: How would the two (into_weak and just weak) be accommodated?

src/address.rs
Restioson: Internal function would have to be in quite a few places but overall might save on comments I guess

src/context.rs
Restioson: Yea it can be not running in the `stopped` and `stopping` methods. Reminds me that I should only return none in stopped because at that point the actor is about to be dropped so any address WILL become invalid, whereas in stopping it can recover.

src/context.rs
Restioson: I doubt it because of the frequency of allocation required (most cases will be 2 allocs per push instead of 1) as well as slower iteration time. I think LinkedList in std lib still has to have sized elements.

src/address.rs
Restioson: A sink implementation might cause conflicting impl errors to arise (two impls have to exist - one for async handlers and one for sync handlers). However it would be possible to return an async fn but would also make it less ergonomic I feel? You'd have to call spawn on that manually. Otherwise I could implement both (like spawn vs create) but then there'd be four methods. Ultimately this is just a convenience wrapper so the flexibility can be achieved just with a few more lines of code using the public API

src/address.rs
gegy1000: `downgrade` and `into_downgraded`

src/address.rs
gegy1000: Imo nicer to go with async fn. Is kinda weird to have library implicitly spawning onto global executor. Async fn is pretty standard anyway. `forward` returns a future, too

Restioson: I looked into this, and it isn't really possible without boxing (trait fns with async are boxed). It's just sugar over the direct thing so I'll leave that as a comment and then users can choose to do that on their own if they want to. It's all built with the public API, anyway.

PR #10, opened 2020-03-17
Remove doc_spotlight feature flag

Emilgardis:
The feature flag has been removed.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,8 +5,7 @@
     weak_counts,
     specialization,
     type_alias_impl_trait,
-    doc_cfg,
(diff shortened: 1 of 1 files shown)


Restioson:
Thanks for the PR :+1: 

What do you think about the implications for the minimum rustc version though?

Emilgardis:
msrv shouldn't change, change was done in rust-lang/rust#69514
This feature is not used in this release of xtra, so this is backwards compatible. This is more of a cleanup + bumping "maximum" supported version.

Restioson:
Released in v0.2.6

PR #12, opened 2020-06-08
wasm-bindgen runtime

stoically:
Thanks for xtra! Exactly what I was looking for.

The README states that xtra is runtime agnostic, but I couldn't find an easy way to use it in WASM without the tokio or async-std runtime feature, hence I added the wasm-bindgen support. Hope this is useful.

To run the example `cargo install wasm-pack` is required.

What's missing in this PR are the methods depending on time, like `notify_later`, but I'd be fine to file that as follow-up if the wasm-bindgen feature generally is something you want to support.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,12 +16,16 @@ futures = { version = "^0.3" }
 async-trait = { version = "^0.1", optional = true }
 tokio = { version = "^0.2", features = ["rt-core", "time"], optional = true }
 async-std = { version = "^1", features = ["unstable"], optional = true }
+wasm-bindgen = { version = "^0.2", optional = true }
+wasm-bindgen-futures = { version = "^0.4", optional = true }
+futures-timer = { version = "3.0.2", optional = true }
 
 [features]
 default = ["stable"]
 stable = ["async-trait"]
 with-tokio-0_2 = ["tokio"]
 with-async_std-1 = ["async-std"]
+with-wasm_bindgen-0_2 = ["wasm-bindgen", "wasm-bindgen-futures", "futures-timer/wasm-bindgen"]
 
 [[example]]
 name = "basic_tokio"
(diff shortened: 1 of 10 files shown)


Restioson:
Hey, thanks for the PR! 

I assume when you say "methods depending on time, like `notify_later`" you mean `notify_interval` and `notify_after`? `notify_later` itself is just later relative to immediate notifications.

Otherwise, this PR looks good, and I am happy to merge :) It will probably be released alongside other additions such as compiling on stable and `smol` support in 0.3 soonish.

stoically:
> I also plan to add methods for smol, now that it has been released.

That's super cool! Also looking forward to the [WASM support in smol](https://github.com/stjepang/smol/pull/99). 

> I assume when you say "methods depending on time, like notify_later" you mean notify_interval and notify_after? notify_later itself is just later relative to immediate notifications.

Yeah, sorry, that's what I meant. Just added support for `notify_interval` and `notify_after` using `futures-timer` with the latest force-push.

Looking forward to the 0.3 release - stable rust support is awesome :)

Restioson:
> Also looking forward to the WASM support in smol.

Wow, that sounds cool! 

> Just added support for notify_interval and notify_after using futures-timer with the latest force-push.

Okay, cool :) I will merge now then.

Restioson:
I've edited some stuff around `wasm-bindgen` in [this commit](https://github.com/Restioson/xtra/pull/14/commits/1d0ac3fbaff39b552f0411c630a7f458d336f29c) to remove default features (for better configurability). The tests still run and pass successfully for me. However, I have limited prior experience with wasm and wouldn't know if I have unwittingly made an error, so if you could look at it too that would be greatly appreciated!

stoically:
Looks good to me, I'll let you know if I encounter an issue. I'm curious, why did you disable default-features for wasm-bindgen but not for the other runtimes?

Restioson:
Different reasons:
- async_std would not compile without it. Turns out they have a `cfg(feature = "default")` on the `task::spawn` fn.
- smol does not have any default features
- and neither does tokio

PR #15, opened 2020-06-10
`Context::handle_while` and `Context::yield_once` for interleaving messages

### LINKED ISSUE #13: Deadlock when sending circular message
When sending async messages like `A -[send]-> B -[send]-> A`, xtra will be stuck in a deadlock. After researching a bit it seems that this is [expected behavior](https://www.reddit.com/r/rust/comments/fbe20b/xtra_a_tiny_fast_and_safe_but_nightlyexperimental/fj4uqpa/). There was `notify` mentioned to mitigate that, but that doesn't seem to work in my case. If it's possible a pointer/example would be helpful.

It'd also nice if it'd be somehow possible to print a warning or something when this happens, because keeping track of who sends messages to whom in an actor system can get quite hard. Going by that reddit thread I can see that it's not an easy problem to solve.

Kudos again for xtra as it is!

Restioson:
This PR adds two methods:
- `Context::handle_while`, allowing the user to handle messages as they arrive while running a given future. This future must be `Unpin` and may not borrow from `self`. It is implemented as a `select` between the two futures, repeating the future waiting for the next message until the first future completes.
- `Context::yield_once`, allowing the user to yield control back to the manager and handle a single message.

This change required some internal refactoring, namely moving handling of messages to the context, but was otherwise very simple to implement and does not introduce much technical debt.

To do:
- [x] Documentation
- [x] Handle immediate notifications first

Potentially closes #13

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,12 @@ path = "examples/nightly.rs"
 required-features = ["with-tokio-0_2", "tokio/full"]
 
 [[example]]
+name = "interleaved_messages"
+path = "examples/interleaved_messages.rs"
+required-features = ["with-tokio-0_2", "tokio/full", "stable"]
+
+
+[[example]]
 name = "crude_bench"
 path = "examples/crude_bench.rs"
 required-features = ["with-tokio-0_2", "tokio/full", "stable"]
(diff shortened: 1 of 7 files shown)


stoically:
> I've included this under examples/interleaved_messages.rs. Is this okay? Shall I attribute you in a comment at the top for writing most of it?

Sure, no problem. I've only modified your example, so no attribution required. Thanks for asking.

PR #26, opened 2021-01-04
Update some dependencies

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>

* update `tokio` to  1.0
* update `flume` to 0.10.1
* rename feature `with-tokio-0_3` to `with-tokio-1`
* rename feature `with-smol-1_1` to `with-smol-1`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,8 +11,8 @@ jobs:
     strategy:
       matrix:
         runtime:
-          - with-tokio-0_3
-          - with-smol-1_1
+          - with-tokio-1
+          - with-smol-1
           - with-async_std-1
           - with-wasm_bindgen-0_2
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,34 +19,34 @@ async-trait = "0.1.36"
 pollster = "0.2"
 catty = "0.1.4"
 barrage = "0.2.1"
-flume = { git = "https://github.com/zesterer/flume", default-features = false, features = ["async"]}
+flume = { version = "0.10.1", default-features = false, features = ["async"] }
 
-tokio = { version = "0.3", features = ["rt", "time"], optional = true }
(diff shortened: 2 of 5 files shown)


koushiro:
@Restioson PTAL

Restioson:
> rename feature with-smol-1_1 to with-smol-1

This cannot be renamed I believe, since smol 1.0 to 1.1 actually had a small breaking change which affected xtra (lifetime added to `Execturo`). We could rename it to imply `smol-greater-than-1_1` maybe?

koushiro:
okay, I will revert the feature name.

koushiro:
@Restioson  If we specify the semver of smol as `1.1` or `1.2` in the `xtra`, then downstream crate will not be able to use smol 1.0 in the same crate. So I think the feature name `with-smol-1` is fine.

Restioson:
Xtra cannot work with smol 1.0 in any case. It must have smol 1.1 to function otherwise it will simply not compile

koushiro:
> Xtra cannot work with smol 1.0 in any case. It must have smol 1.1 to function otherwise it will simply not compile

yeah I know, the feature `with-smol-1` is just a name. 
Since we have specified the semver as `1.2` in Cargo.toml, the downstream crate can obviously only use soml 1.2 or newer version

Restioson:
I think its better to just support the minimum version that we can, in case they want to use 1.1 instead of 1.2. I think we should keep the feature name with-smol-1_1 to reflect that

koushiro:
> I think its better to just support the minimum version that we can, 

I agree with it.

But to be honest, I think `_1` is completely unnecessary．If  you want it, why not `with-async_std-1_0`.

Restioson:
It's not `with-async_std-1_0` because it's correct in the cargo.toml to specify async_std as `async_std = "1"`. The `0` can be left off here. It can't be left off with `1_1`.

I will merge it now, thanks so much!

PR #32, opened 2021-01-12
Add spawn_default to the ActorManager

mneumann:
In addition to that, impl Default for each runtime's Spawner.

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -56,6 +56,13 @@ impl<A: Actor> ActorManager<A> {
         addr
     }
 
+    /// Spawn the actor's main loop on the given runtime using the default spawner. This will allow
+    /// it to handle messages.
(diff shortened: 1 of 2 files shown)


src/manager.rs
```diff
@@ -56,6 +56,13 @@ impl<A: Actor> ActorManager<A> {
         addr
     }
 
+    /// Spawn the actor's main loop on the given runtime using the default spawner. This will allow
+    /// it to handle messages.
+    pub fn spawn_default<S: Spawner + Default>(self) -> Address<A> {
```
Restioson, line 61: This is not quite the same as with your original issue, right?

mneumann: Not really. Probably your extension trait suggestion `spawn_global` works better.

Restioson: Ok. I will try to get a basic implementation of it tomorrow.

PR #33, opened 2021-04-16
Quality of my life

qm3ster:
* Take `Actor` by value in `stopped` lifecycle method (API breaking, but more permissive)
* Implement `PartialEq + Eq + PartialOrd + Ord + Hash` on `Address` by pointer identity
  (uses the `ref_counter`, because `flume::Sender` doesn't expose `Arc` or pointer)

---
### Original PR description:

real find and replace hours

### DIFF
--- a/src/refcount.rs
+++ b/src/refcount.rs
@@ -61,6 +61,9 @@ pub trait RefCounter: Clone + Unpin + Send + Sync + 'static {
 
     #[doc(hidden)]
     fn into_either(self) -> Either;
+
+    #[doc(hidden)]
+    fn as_ptr(&self) -> *const AtomicBool;
 }
 
 impl RefCounter for Strong {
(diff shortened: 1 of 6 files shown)


qm3ster:
Now with raisins:tm:

src/refcount.rs
```diff
@@ -79,6 +82,10 @@ impl RefCounter for Strong {
     fn into_either(self) -> Either {
         Either::Strong(self)
     }
+
+    fn as_ptr(&self) -> *const AtomicBool {
```
Restioson, line 86: Not sure about having this in the RefCounter trait, this would make it public API, no?

qm3ster: It's `#[doc(hidden)]` in the trait declaration of this already-for-internal-use trait.
Seemed less messy this way.
If we want people to implement more kinds of `RefCounter` themselves 😬 this could be a separate trait, so they could choose if their new `RefCounter` will participate in pointer shenanigans.

Restioson: Fair enough, it would have to be a pub trait anyway.

PR #39, opened 2021-09-24
Annotate SendFuture with #[must_use] attribute

klochowicz:
Prevents misuses of the API, e.g.

`address.send()` without calling `.await` on it.


Restioson:
Thanks!

PR #40, opened 2021-09-27
Annotate DoSendFuture with #[must_use] attribute

klochowicz:
Prevents misuses of the API, e.g.

`address.do_send_async()` without calling `.await` on it.

Sorry, I missed the second use-case - ideally I would have sent a single PR!


Restioson:
Thanks and no problem! I will cut a release with this included when I have time. I am a bit busy as the university term has begun again but I hope to cut a release this weakened.

PR #42, opened 2021-10-05
Remove `Actor` trait bounds from structs

thomaseizinger:
Trait bounds that are specified on structs ripple easily through
entire codebases.
In this particular case, requiring `Actor` for an `Address` is
unnecessary and makes it harder to hold on `Address` to an actor
that is represented by a type parameter.

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -116,7 +116,7 @@ impl Error for Disconnected {}
 /// should be used instead. An address is created by calling the
 /// [`Actor::create`](../trait.Actor.html#method.create) or
 /// [`Context::run`](../struct.Context.html#method.run) methods, or by cloning another `Address`.
-pub struct Address<A: Actor, Rc: RefCounter = Strong> {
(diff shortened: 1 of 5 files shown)


Restioson:
Thanks for the PR! I wonder if `Rc: RefCounter` bounds can also be removed in some cases?

thomaseizinger:
I can try. Are you okay if I do that in a follow-up PR?

thomaseizinger:
> Thanks for the PR! I wonder if `Rc: RefCounter` bounds can also be removed in some cases?

We can't because it is required in `Drop` and every bound on `Drop` needs to be present in the struct definition.

Restioson:
Ah, I see, makes sense.

dvdplm:
Is there a release including this on the horizon? Rough timeline? :)

Restioson:
Hi, I can look at cutting a release after I've finished writing my exams which should be by the end of this month - for now if you need this the best option is to pin to a git commit.

PR #47, opened 2021-11-24
Mark message_channel::SendFuture as must_use

luckysori:
This way the compiler will remind us to `await` this kind of future.

It's the same QoL improvement which was introduced for `Address` in https://github.com/Restioson/xtra/pull/39.


Restioson:
Thanks!

PR #48, opened 2021-11-26
Improved handling of attached streams and delayed notifications

rrauch:
First of all I would like to thank you for `xtra`. Playing around with it has been a breeze.
I have noticed though that, currently, when using `Address::attach_stream` `stream.next` is waited upon even after the address becomes disconnected. In my particular use-case this has lead to `attach_stream` simply becoming stuck as no new messages are added to the stream.

To fix this I had to make some changes - not only to `attach_stream`, but also to `Address` & `Context` to track the status of all contexts attached to one address and to get notified as soon as they drop. After my changes `attach_stream` returns immediatley once the address has become disconnected. Unrelated to this, I added `Address::join` to improve ergonomics when needing to wait for the completion of an actor.

I had to add the [stream_cancel](https://docs.rs/stream-cancel/) crate as a dependency to deal with stream cancellation and  drop notification. It's pretty lightweight and works well and the `Tripwire` & `Trigger` synchronization primitives are very useful. Please let me know if you want something else instead. I looked through the existing deps and didn't see anything that could be used for this.

I've also made some changes to `Context::notify_interval` & `Context::notify_after`. Firstly, I added code to automatically stop those background tasks once the context dropped. I also noticed that these background tasks hold strong addresses and could therefore prevent an actor from being stopped. I don't believe this was the intended behaviour, so I added a `downgrade` to make sure the task holds a weak address instread.

Let me know how you feel about these changes. Cheers.


Restioson:
Hi, thanks for taking a look at this! I'll review it shortly

src/address.rs
```diff
@@ -119,6 +120,10 @@ impl Error for Disconnected {}
 pub struct Address<A, Rc: RefCounter = Strong> {
     pub(crate) sender: Sender<AddressMessage<A>>,
     pub(crate) ref_counter: Rc,
+    /// Activates once all contexts on this address have been dropped.
+    /// Used in [`Address::attach_stream`](struct.Address.html#method.attach_stream) to stop waiting
+    /// for new messages when this `Address` becomes disconnected and in [`Address::join`].
+    pub(crate) tripwire: Tripwire
```
Restioson, line 126: Not sure whether adding this to the address itself is a good idea, since it would inflate the size of the address

Restioson (commented):
I think this functionality could be moved into the `Rc` struct or something, with a new `RcInner` type that is stored behind the Arc, to avoid making the address bigger than it already is. However, I also wonder if bringing in the `stream-cancel` crate is the right decision. I'm not sure if I'm maybe looking at this wrong, but it looks like it may bring in [tokio](https://github.com/jonhoo/stream-cancel/blob/master/Cargo.toml#L22), which is an additional dependency I'd rather avoid. It shouldn't be too complicated to implement in xtra, though.

Restioson:
The `event-listener` crate might be a suitable alternative - see [EventListener](https://docs.rs/event-listener/2.5.1/event_listener/struct.EventListener.html). The main upside would be that it is already in the dependency graph of xtra, as it is a dependency of `barrage`. Let me know what you think

src/drop_notice.rs
```diff
@@ -0,0 +1,80 @@
...
+            self.listener = Some(listener);
+        } else {
+            drop(event);
+        }
+
+        let listener = self.listener.as_mut().unwrap();
+        futures_util::pin_mut!(listener);
+        listener.poll(cx)
+    }
+}
```
Restioson, line 80: nit: missing newline

src/address.rs
```diff
@@ -120,10 +120,10 @@ impl Error for Disconnected {}
 pub struct Address<A, Rc: RefCounter = Strong> {
     pub(crate) sender: Sender<AddressMessage<A>>,
     pub(crate) ref_counter: Rc,
-    /// Activates once all contexts on this address have been dropped.
+    /// Resolves once all contexts on this address have been dropped.
     /// Used in [`Address::attach_stream`](struct.Address.html#method.attach_stream) to stop waiting
     /// for new messages when this `Address` becomes disconnected and in [`Address::join`].
-    pub(crate) tripwire: Tripwire
+    pub(crate) stop_notice: DropNotice,
```
Restioson, line 126: This will also lead to the address becoming bigger - do you think it'd be possible to store the event inside the DropNotice directly (i.e not behind an Arc), and then store the DropNotice inside the ref_counter's allocation, which is already an Arc?

One possible issue with this is that to wait on the stop notice, it would need to upgrade, which increases the strong count, so that alone would inhibit the actor being dropped. So, if that's done (which is probably better given that we want to keep the Address struct as small as it can be because there are usually a lot of them), we would need to use a separate counter for the number of actors, rather than reusing strong_count

rrauch:
I have moved the `DropNotice` to the `RefCounter`, inside the existing Arc as you suggested. It's definiely much leaner this way and `Address` does not become bigger.
As you mentioned, in the `Weak` variant of the `RefCounter` `upgrade` has to be called in order to access the Arc, but only to clone the `DropNotice` held inside, not for the whole wait duration. The upgraded Arc is dropped again immediately, so I don't think this will negatively affect the accuracy of the strong count. In case the upgrade returns `None`, a dummy `DropNotice` is returned that resolves immediately - which correctly reflects the `disconnected`-state of the `Address`.

About the design of the `DropNotifier`/`DropNotice` pair: the reason the `DropNotice` needs a (weak) reference to the `Event` in the `DropNotifier` is that `DropNotice` needs to be `Clone`, in order to use multiple instances independently (in different tasks / possibly different threads). Unfortunately, `EventListener` isn't `Clone`, so there is the need to keep a reference to `Event` and call `Event::listen()` instead. I don't see how this could be achieved otherwise when building upon `event_listener`.

I've rewritten `DropNotice` a bit, and it will now attempt to access the `Event` at-most once, and drops the reference asap. It can even work without an event at all - this is useful in the "dummy"-case mentioned above.
Right now, I can't see what else could be done. At this point I'm quite happy with the result.
Looking forward to your feedback.

src/refcount.rs
```diff
@@ -113,9 +122,16 @@ impl RefCounter for Weak {
...
     }
 
-    fn as_ptr(&self) -> *const AtomicBool {
+    fn as_ptr(&self) -> *const (AtomicBool, DropNotice) {
         ArcWeak::as_ptr(&self.0)
     }
+
+    fn disconnect_notice(&self) -> DropNotice {
+        match self.0.upgrade() {
+            Some(arc) =>  arc.1.clone(),
```
Restioson, line 131: nit: unnecessary space here. It's probably just best to run the whole PR through rustfmt

src/context.rs
```diff
@@ -131,7 +144,7 @@ impl<A: Actor> Context<A> {
     /// Stop all actors on this address
     fn stop_all(&mut self) {
         if let Some(strong) = self.ref_counter.upgrade() {
-            strong.0.store(false, Ordering::Release)
+            strong.0.0.store(false, Ordering::Release)
```
Restioson, line 147: Shouldn't this (and probably somewhere else too) call shared_drop_notifier and set it as dropped? Otherwise, am I just missing where it's called here?

Restioson:
> The upgraded Arc is dropped again immediately, so I don't think this will negatively affect the accuracy of the strong count

Okay, awesome! I think this is a lot better then than I thought it would be. However, I think there would still exist the following race condition:

1. There is one strong address and one weak address. Strong count is 1
2. The thread with the weak address calls `join`. Strong count is incremented to 2.
3. Before it has had a chance to drop the upgraded Arc, the strong address is dropped. Since it's 2, it does not send the last address notification.
4. The actor will now never close.

I think this could be rectified by checking the strong_count on the address after the notifier has been cloned (and imperatively, after the upgraded arc has been dropped!), and then sending the last_address notification if it's equal to 0.

Restioson (commented):
Looks good aside from the thing I mentioned, as to where the drop notifier is called. In general this PR could use a few test cases, but I'd also be happy to add them after the fact

src/context.rs
rrauch: The `DropNotifier` gets triggered when it's dropped - hence the name.
So once all references to the `shared_drop_notifier` are gone, the `DropNotifier` is dropped and the corresponding `DropNotice`s get notified automatically - no need to call stop or anything. I find this approach much safer & easier to use. There's no way one could forget to call it and no need to keep an extra counter or anything.

rrauch:
> > The upgraded Arc is dropped again immediately, so I don't think this will negatively affect the accuracy of the strong count
> 
> Okay, awesome! I think this is a lot better then than I thought it would be, actually - I thought the fix would involve a separate counter sweat_smile This is a lot more elegant. However, I think there would still exist the following race condition:
> 
> 1. There is one strong address and one weak address. Strong count is 1
> 2. The thread with the weak address calls `join`. Strong count is incremented to 2.
> 3. Before it has had a chance to drop the upgraded Arc, the strong address is dropped. Since it's 2, it does not send the last address notification.
> 4. The actor will now never close.
> 
> I think this could be rectified by checking the strong_count on the address after the notifier has been cloned (and imperatively, after the upgraded arc has been dropped!), and then sending the last_address notification if it's equal to 0.

Hmm, good point. I think this race condition is possible.
I even think the current codebase might actually be vulnerable to it *already*:

`Weak::is_connected` does exactly the same thing - it briefly upgrades the Arc without any extra checking.
If I'm not mistaken the scenario you described could already happen when `is_connected` is called on a weak address.

To fix this, sending the `last_address` notification after explicitly dropping the arc and checking the strong_count would be an option, as in your suggestion, but I think it would be a bit messy.
Firstly, `RefCounter` would need access to a `Sender`, and adding it could be messy - especially as `Sender` is generified.
Secondly, it kinda breaks encapsulation of responsibility. So far, `RefCounter` has a clear role and is self contained. If it now also has to send messages to actors in certain circumstances, it would make it harder to reason about in my opinion.

I've looked at several other options and I think adding a `RwLock` to `RefCounter` would fix the race condition and would be a cleaner approach overall.
Specifially, I'm suggesting the following changes:

1. Add and `Arc<RwLock>` to `Strong` and `Weak`, but outside the existing Arc of course.
2. Set the write lock at the beginning of `Weak::is_connected` & `Weak::disconnect_notice`, before attempting to upgrade the Arc. The lock will remaining locked until the end of the function. Here it's important to explicitly drop the upgraded Arc to avoid the race condition!
3. Aquire a read lock at the beginning of `Strong::is_last_strong` and possibly also at `Strong::strong_count` & `Weak::strong_count`. Now these functions will always return the accurate value.

I think using a lock is the best option in this case. It's not ideal as it adds one more Arc to the RefCounter, but I see no way to avoid this race condition without adding *something* to the RefCounter anyway, so it might just be an `Arc<RwLock>`.
It's important to use the lock sparingly to not noticeably affect performance. That's why I'm suggesting not locking the whole existing Arc and just use it in these few functions. I've ran the included benchmarks a couple of times and I can't see any measurable differences with the new `RwLock`.

Let me know if this works for you.

src/context.rs
Restioson: Oh, cool, I missed that, sorry!

rrauch:
@Restioson I went ahead and implemented the changes I suggested in my comment yesterday. This should fix the race condition while keeping things sane.
Running the benchmarks didn't show any measurable difference, so the new lock doesn't have any real impact on performance.

I also let `rustfmt` reformat the code and added some tests, as you suggested. I think this PR is ready for merging.

src/refcount.rs
```diff
@@ -8,22 +10,35 @@ use crate::private::Sealed;
 /// out more.
 #[derive(Clone)]
 // TODO AtomicBool for disconnected when forcibly stopped ?
-pub struct Strong(pub(crate) Arc<AtomicBool>);
+// The RwLock is there to prevent exposing a temporarily inconsistent strong_count caused by brief
+// Arc::upgrade calls in some `Weak` functions below. If exposed, it could lead to a race condition
+// that can prevent an Actor from being stopped.
+pub struct Strong(pub(crate) Arc<(AtomicBool, DropNotice)>, Arc<RwLock<()>>);
```
Restioson, line 16: This unfortunately kinda brings us back to a larger address struct. Maybe it's just a hit we need to take though.

src/refcount.rs
```diff
@@ -65,39 +80,55 @@ pub trait RefCounter: Sealed + Clone + Unpin + Send + Sync + 'static {
...
     }
+
+    fn disconnect_notice(&self) -> DropNotice {
+        self.0 .1.clone()
+    }
 }
 
 impl RefCounter for Weak {
     fn is_connected(&self) -> bool {
+        let lock = self.1.write().unwrap();
```
Restioson, line 119: I wonder if it'd be possible to just use a normal bool that would be inside of the RwLock? Or would that also be susceptible to the same race condition?

tests/basic.rs
```diff
@@ -111,3 +113,39 @@ async fn test_stop_and_drop() {
     smol::spawn(fut).await;
     assert_eq!(drop_count.load(Ordering::SeqCst), 3);
 }
+
+struct StreamCancelMessage;
```
Restioson, line 117: Thanks for this test :)

Restioson:
Looks like CI is failing on rustfmt here

src/context.rs
```diff
@@ -144,7 +144,7 @@ impl<A: Actor> Context<A> {
     /// Stop all actors on this address
     fn stop_all(&mut self) {
         if let Some(strong) = self.ref_counter.upgrade() {
-            strong.0.0.store(false, Ordering::Release)
+            strong.0 .0.store(false, Ordering::Release)
```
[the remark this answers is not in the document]
rrauch: Yup, that's rustfmts doing.

src/refcount.rs
```diff
@@ -8,22 +10,35 @@ use crate::private::Sealed;
...
 // TODO AtomicBool for disconnected when forcibly stopped ?
-pub struct Strong(pub(crate) Arc<AtomicBool>);
+// The RwLock is there to prevent exposing a temporarily inconsistent strong_count caused by brief
+// Arc::upgrade calls in some `Weak` functions below. If exposed, it could lead to a race condition
+// that can prevent an Actor from being stopped.
+pub struct Strong(pub(crate) Arc<(AtomicBool, DropNotice)>, Arc<RwLock<()>>);
 
 impl Strong {
+    pub(crate) fn new(connected: AtomicBool, drop_notice: DropNotice) -> Self {
+        Self {
```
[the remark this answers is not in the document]
rrauch: Will do

src/refcount.rs
rrauch: I don't see how this can be avoided I'm afraid.
The `RwLock` has to be outside the existing `Arc`, otherwise it would be pointless. 
On 64 bit, this changes the size of `Address` from 16 to 24 bytes, so an 8 byte increase unfortunately.
However, with the extra size comes correctness, so not too bad imho. I'm open to suggestions of course.

src/refcount.rs
rrauch: Not sure I follow. The `RwLock` actually only contains a `()`, not the main counting `Arc`.

rrauch:
> Looks like CI is failing on rustfmt here. It seems like it doesn't run the tests if the formatting is wrong, either.

@Restioson I think CI is failing because `rustfmt` is complaining about recent changes in `master`, specifically the tracing related changes. I get the same error when running `rustfmt` against `master`. Does the CI work with `master`?

src/context.rs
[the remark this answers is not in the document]
Restioson: Alright, that's fine then. So long as the formatting is consistent

src/refcount.rs
Restioson: That's true - it might just be a hit we need to take. I'm happy with this strategy now with the increase, and if someone else in the future wants to reduce the size because they need to for whatever reason, they can go for it and I'll merge that too.

src/refcount.rs
Restioson: It's probably not important - I don't really remember what I meant with this.

Restioson:
I think you're right - could you verify that tests pass manually then? Otherwise, happy to merge. Sorry this took so long!

rrauch:
> I think you're right - could you verify that tests pass manually then?

Yes, everything passes in my branch. And no complaining from `rustfmt` or clippy.

> I'll fix the formatting on main when I get back from holiday.

I've went ahead and fixed it in #50. I also found an issue with a missing feature flag that causes `master` not to compile when `with-tracing-0_1` isn't enabled.

> Sorry this took so long!

No worries at all! I appreciate your responsiveness and patience. I'm still getting my bearings with Rust - and it shows!
Enjoy your holidays!

Issue #41, opened 2021-10-03
Allow actors to `poll` other futures as part of their event loop?


thomaseizinger:
I am trying to integrate `xtra` with `libp2p`.

`libp2p`'s main design is a big state machine that is called a `Swarm`. I has a cancellation-safe, async `next` method but also a manual `poll_next_unpin` function that advances the internal state.

Calling this function is necessary to have the network layer (represented by the `Swarm`) make progress. Additionally, in order to interact with the `Swarm`, one needs a mutable reference.

It would be nice if `xtra` could expose a way of embedding another async component like `Swarm` that could be polled as part of its event loop. Polling the `Swarm` yields events which I'd like to handle within the actor, probably as notifications to itself?

`attach_stream` is not an option because I still need a &mut reference to the `Swarm` whenever I want to interact with the network layer.

The only way of integrating `xtra` with `libp2p` at the moment is to `Arc<Mutex>` the `Swarm`. This works but it would be nice if there were a more elegant solution!

Thoughts?

Restioson:
Hm, as far as the event loop of the actor goes, it lives in [`Context::run`](https://github.com/Restioson/xtra/blob/master/src/context.rs#L191). The main question to resolve, I guess, would be an elegant way to integrate this loop with the other future to run. Two other possible methods of integration would be [`Context::handle_while`](https://github.com/Restioson/xtra/blob/master/src/context.rs#L320) and [`Context::yield_once`](https://github.com/Restioson/xtra/blob/master/src/context.rs#L302). I wonder if the main `Context::run` loop could be replaced by a custom `Context::handle_while` loop? Something like

```rust
let actor = MyActor;
let (addr, ctx) = Context::new(None);

loop {
    ctx.handle_while(actor, swarm.next()).await
}
```

However, if the actor needs a mutable reference to the swarm, I'm not sure how this could work in any case. It basically would mean that instead of selecting over the two futures, you would need to run one and then the other. Something like

```rust
let actor = MyActor { swarm };
let (addr, ctx) = Context::new(None);

loop {
    ctx.yield_once(actor).await;
    actor.swarm.next().await;
}
```

[...]

Issue #49, opened 2021-11-30
Have `xtra` detect deadlocks


thomaseizinger:
It would be nice if `xtra` could detect deadlocks where two actors use `.send` to invoke each other from their handlers. Perhaps we can implement some form of "delayed" log message that triggers after a handler has been processing for longer than, say, 5 seconds? If the message itself is also logged, that would help a lot in finding these kind of bugs!

Restioson:
I'm unfamiliar with the techniques used in deadlock detection, but I know parking_lot has experimental deadlock detection. So, if someone else has an implementation strategy, I'd be happy to look at merging a PR.

thomaseizinger:
Are you interested in something like this: https://github.com/comit-network/xtra/commit/15fdd524f2ed3ac94acfc92e21d5f285a485c23d?

Restioson:
Essentially, rather than true deadlock detection, logging if one message has been handled for a long time?

thomaseizinger:
> Essentially, rather than true deadlock detection, logging if one message has been handled for a long time?

Correct. The inspiration came from `cargo test`, which outputs a log if a test takes longer than 60 seconds.

thomaseizinger:
I think so too yeah.

Let's close this in favour of pushing for a more modular design!

PR #51, opened 2021-12-19
Refactor context

thomaseizinger:
See commit messages for details.

Some of this is inspired by the discussion in #41.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -16,6 +16,11 @@ use crate::envelope::{MessageEnvelope, NonReturningEnvelope};
 use crate::manager::{AddressMessage, BroadcastMessage, ContinueManageLoop};
 use crate::refcount::{RefCounter, Strong, Weak};
 use crate::{Actor, Address, Handler, KeepRunning, Message};
+use futures_core::FusedFuture;
+use futures_util::__private::async_await;
+use std::pin::Pin;
+use std::task;
+use task::Poll;
 
 /// `Context` is used to control how the actor is managed and to get the actor's address from inside
 /// of a message handler.
@@ -224,16 +229,68 @@ impl<A: Actor> Context<A> {
         let mut broadcast_recv = broadcast_rx.recv_async().fuse();
 
         loop {
-            let msg = futures_util::select! {
-                msg = addr_recv => {
-                    addr_recv = addr_rx.recv_async().fuse();
-
-                    Either::Right(msg.unwrap())
+            let msg = {
+                enum __PrivResult<_0, _1> {
+                    _0(_0),
+                    _1(_1),
                 }
-                msg = broadcast_recv => {
-                    broadcast_recv = broadcast_rx.recv_async().fuse();
+                let __select_result = {
+                    let mut __poll_fn = |__cx: &mut task::Context<'_>| {
+                        let mut __any_polled = false;
+                        let mut _0 = |__cx: &mut task::Context<'_>| {
+                            let mut addr_recv = unsafe { Pin::new_unchecked(&mut addr_recv) };
+                            if FusedFuture::is_terminated(&addr_recv) {
+                                None
+                            } else {
+                                Some(
+                                    future::FutureExt::poll_unpin(&mut addr_recv, __cx)
+                                        .map(__PrivResult::_0),
+                                )
+                            }
+                        };
+                        let _0: &mut dyn FnMut(&mut task::Context<'_>) -> Option<Poll<_>> = &mut _0;
+                        let mut _1 = |__cx: &mut task::Context<'_>| {
+                            let mut broadcast_recv =
+                                unsafe { Pin::new_unchecked(&mut broadcast_recv) };
+                            if FusedFuture::is_terminated(&broadcast_recv) {
+                                None
+                            } else {
+                                Some(
+                                    future::FutureExt::poll_unpin(&mut broadcast_recv, __cx)
+                                        .map(__PrivResult::_1),
+                                )
+                            }
+                        };
+                        let _1: &mut dyn FnMut(&mut task::Context<'_>) -> Option<Poll<_>> = &mut _1;
+                        let mut __select_arr = [_0, _1];
+                        async_await::shuffle(&mut __select_arr);
 
-                    Either::Left(msg.unwrap())
+                        for poller in &mut __select_arr {
+                            match poller(__cx) {
+                                Some(x @ Poll::Ready(_)) => return x,
+                                Some(Poll::Pending) => {
+                                    __any_polled = true;
+                                }
+                                None => {}
+                            }
+                        }
+                        if !__any_polled {
+                            panic!("all futures in select! were completed, but no `complete =>` handler was provided");
+                        } else {
+                            Poll::Pending
+                        }
+                    };
+                    future::poll_fn(__poll_fn).await
+                };
+                match __select_result {
+                    __PrivResult::_0(msg) => {
+                        addr_recv = addr_rx.recv_async().fuse();
+                        Either::Right(msg.unwrap())
+                    }
+                    __PrivResult::_1(msg) => {
+                        broadcast_recv = broadcast_rx.recv_async().fuse();
+                        Either::Left(msg.unwrap())
+                    }
                 }
             };
 
--- a/src/context.rs
+++ b/src/context.rs
@@ -296,17 +296,15 @@ impl<A: Actor> Context<A> {
     }
 
     /// This is a combinator to avoid holding !Sync references across await points
-    fn recv_once(
-        &self,
-    ) -> impl Future<Output = Either<BroadcastMessage<A>, AddressMessage<A>>> + '_ {
-        future::select(
-            self.broadcast_receiver.recv_async(),
-            self.receiver.recv_async(),
-        )
-        .map(|next| match next {
-            Either::Left((res, _)) => Either::Left(res.unwrap()),
-            Either::Right((res, _)) => Either::Right(res.unwrap()),
-        })
+    async fn recv_once(&self) -> Either<BroadcastMessage<A>, AddressMessage<A>> {
+        futures_util::select! {
+            msg = self.broadcast_receiver.recv_async().fuse() => {
+                Either::Left(msg.unwrap())
+            }
+            msg = self.receiver.recv_async().fuse() => {
+                Either::Right(msg.unwrap())
+            }
+        }
     }
 
     /// Yields to the manager to handle one message.
--- a/src/context.rs
+++ b/src/context.rs
@@ -296,7 +296,7 @@ impl<A: Actor> Context<A> {
         let mut inbox = self.inbox();
 
         loop {
-            futures_util::select_biased! {
+            futures_util::select! {
                 msg = inbox.select_next_some() => {
                     self.tick(msg, actor).await;
                 }
--- a/src/context.rs
+++ b/src/context.rs
@@ -5,8 +5,8 @@ use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 
 use flume::{Receiver, Sender};
+use futures_core::stream::BoxStream;
 use futures_util::future::{self, Either};
-use futures_util::stream::FusedStream;
 use futures_util::FutureExt;
 use futures_util::StreamExt;
 
@@ -232,12 +232,8 @@ impl<A: Actor> Context<A> {
 
     /// Handle a message and immediate notifications, returning whether to exit from the manage loop
     /// or not.
-    async fn tick(
-        &mut self,
-        msg: Either<BroadcastMessage<A>, AddressMessage<A>>,
-        actor: &mut A,
-    ) -> ContinueManageLoop {
-        match msg {
+    async fn tick(&mut self, msg: InboxMessage<A>, actor: &mut A) -> ContinueManageLoop {
+        match msg.inner {
             Either::Left(BroadcastMessage::Message(msg)) => msg.handle(actor, self).await,
             Either::Right(AddressMessage::Message(msg)) => msg.handle(actor, self).await,
             Either::Left(BroadcastMessage::Shutdown) => {
@@ -270,9 +266,9 @@ impl<A: Actor> Context<A> {
             return;
         }
 
-        let msg = self.inbox().select_next_some().await;
-
-        self.tick(msg, act).await;
+        if let Some(msg) = self.inbox().next().await {
+            self.tick(msg, act).await;
+        }
     }
 
     /// Handle any incoming messages for the actor while running a given future.
@@ -296,9 +292,16 @@ impl<A: Actor> Context<A> {
         let mut inbox = self.inbox();
 
         loop {
+            let inbox_msg = inbox.next().fuse();
+            futures_util::pin_mut!(inbox_msg);
+
             futures_util::select! {
-                msg = inbox.select_next_some() => {
-                    self.tick(msg, actor).await;
+                msg = inbox_msg => {
+                    if let Some(msg) = msg {
+                        self.tick(msg, actor).await;
+                    }
+
+                    // On `None`, the actor's inbox is closed (shutdown?) but our `fut` hasn't resolved yet. Essentially, we now just wait for `fut` to complete.
                 }
                 result = fut => {
                     return result
@@ -399,24 +402,57 @@ impl<A: Actor> Context<A> {
         Ok(fut)
     }
 
-    fn inbox(&self) -> impl FusedStream<Item = Either<BroadcastMessage<A>, AddressMessage<A>>> {
-        let addr_rx = self.receiver.clone();
-        let broadcast_rx = self.broadcast_receiver.clone();
+    fn inbox(&self) -> Inbox<A> {
+        Inbox::new(self.receiver.clone(), self.broadcast_receiver.clone())
+    }
+}
 
+struct Inbox<A> {
+    inner: BoxStream<'static, Either<BroadcastMessage<A>, AddressMessage<A>>>,
+}
+
+impl<A> Inbox<A>
+where
+    A: 'static,
+{
+    fn new(
+        address_receiver: Receiver<AddressMessage<A>>,
+        broadcast_receiver: barrage::SharedReceiver<BroadcastMessage<A>>,
+    ) -> Self {
         let broadcast_stream =
-            futures_util::stream::unfold(broadcast_rx, |broadcast_rx| async move {
-                match broadcast_rx.recv_async().await {
-                    Ok(msg) => Some((msg, broadcast_rx)),
+            futures_util::stream::unfold(broadcast_receiver, |broadcast_receiver| async move {
+                match broadcast_receiver.recv_async().await {
+                    Ok(msg) => Some((msg, broadcast_receiver)),
                     Err(_) => None,
                 }
             })
             .map(Either::Left)
             .boxed();
 
-        let address_stream = addr_rx.into_stream().map(Either::Right).boxed();
+        let address_stream =
+            futures_util::stream::unfold(address_receiver, |address_receiver| async move {
+                match address_receiver.recv_async().await {
+                    Ok(msg) => Some((msg, address_receiver)),
+                    Err(_) => None,
+                }
+            })
+            .map(Either::Right)
+            .boxed();
 
-        futures_util::stream_select!(broadcast_stream, address_stream).fuse()
+        Self {
+            inner: futures_util::stream_select!(broadcast_stream, address_stream).boxed(),
+        }
     }
+
+    pub async fn next(&mut self) -> Option<InboxMessage<A>> {
+        let inner = self.inner.next().await?;
+
+        Some(InboxMessage { inner })
+    }
+}
+
+struct InboxMessage<A> {
+    inner: Either<BroadcastMessage<A>, AddressMessage<A>>,
 }
 
 /// If and how to continue the manage loop
--- a/src/context.rs
+++ b/src/context.rs
@@ -220,35 +220,22 @@ impl<A: Actor> Context<A> {
         let addr_rx = self.receiver.clone();
         let broadcast_rx = self.broadcast_receiver.clone();
 
-        let mut addr_recv = addr_rx.recv_async();
-        let mut broadcast_recv = broadcast_rx.recv_async();
+        let mut addr_recv = addr_rx.recv_async().fuse();
+        let mut broadcast_recv = broadcast_rx.recv_async().fuse();
 
         loop {
-            let next = future::select(addr_recv, broadcast_recv).await;
+            let msg = futures_util::select! {
+                msg = addr_recv => {
+                    addr_recv = addr_rx.recv_async().fuse();
 
-            let msg = match next {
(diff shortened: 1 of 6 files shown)


src/context.rs
```diff
@@ -221,35 +220,22 @@ impl<A: Actor> Context<A> {
         let addr_rx = self.receiver.clone();
         let broadcast_rx = self.broadcast_receiver.clone();
 
-        let mut addr_recv = addr_rx.recv_async();
-        let mut broadcast_recv = broadcast_rx.recv_async();
+        let mut addr_recv = addr_rx.recv_async().fuse();
```
Restioson, line 223: I assume these fuses are necessary for the futures_util::select macro?

Restioson (approved):
Thanks for the PR!

Broadly, I like these changes. I just want to look at how it changes the performance, for interest's sake (I'd assume minor wins or no change)

Restioson:
Branch:
```
address do_send (ZST message) avg time of processing: 177ns
address do_send (8-byte message) avg time of processing: 264ns
address do_send 2 workers (ZST message) avg time of processing: 261ns
address do_send 2 workers (8-byte message) avg time of processing: 291ns
channel do_send (ZST message) avg time of processing: 191ns
channel do_send (8-byte message) avg time of processing: 229ns
channel send (ZST message) avg time of processing: 337ns
channel send (8-byte message) avg time of processing: 385ns
```

Master:
```
address do_send (ZST message) avg time of processing: 123ns
address do_send (8-byte message) avg time of processing: 213ns
address do_send 2 workers (ZST message) avg time of processing: 120ns
address do_send 2 workers (8-byte message) avg time of processing: 195ns
channel do_send (ZST message) avg time of processing: 113ns
channel do_send (8-byte message) avg time of processing: 232ns
channel send (ZST message) avg time of processing: 311ns
channel send (8-byte message) avg time of processing: 382ns
```

It seems to be somewhat reproducible - I'm not exactly what sure could be causing the performance regressions. Now may be a good time for me to look into implementing criterion benchmarks for xtra to control the execution conditions more tightly.

Restioson:
Check out the criterion branch; I've added a very simple throughput test of sending 100 increment messages and timing how long it takes to process them. On master, seems to be around 170ns per message when batched in 100.

src/context.rs
thomaseizinger: Yes, they need to be `fused` for `select` to work because it might happen that they are polled after completion.

thomaseizinger:
> Broadly, I like these changes. I just want to look at how it changes the performance, for interest's sake (I'd assume minor wins or no change)

Are you squash merging in this repo? Then I would just leave the two commits with the revert in for now.

thomaseizinger:
> Check out the criterion branch; I've added a very simple throughput test of sending 100 increment messages and timing how long it takes to process them.

I've had a look and I am not sure if we are using criterion correctly there? The `setup_fn` is not benchmarked if I understand correctly, only the final `Finish` message is. Is that on purpose?

I am working on a patch on top of your branch that makes use of criterion's async bench support. I am quite the benchmark noob so we will see how far I get there :D

thomaseizinger:
Here are my results:

## Criterion branch:

```
send_zst/1              time:   [8.1708 us 8.2691 us 8.3727 us]
                        change: [-9.2457% -7.7199% -6.2662%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  2 (2.00%) low mild
  4 (4.00%) high mild
  2 (2.00%) high severe
send_zst/10             time:   [84.130 us 84.874 us 85.558 us]
                        change: [-1.9468% -0.4509% +1.0296%] (p = 0.56 > 0.05)
                        No change in performance detected.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe
send_zst/100            time:   [843.49 us 851.96 us 860.68 us]
                        change: [-7.0818% -5.1565% -3.2959%] (p = 0.00 < 0.05)
                        Performance has improved.
send_zst/1000           time:   [8.4739 ms 8.5376 ms 8.6014 ms]
                        change: [-1.6435% -0.3651% +0.8715%] (p = 0.58 > 0.05)
                        No change in performance detected.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
```

## This branch:

```
send_zst/1              time:   [8.5991 us 8.7618 us 8.9253 us]
                        change: [+2.4803% +3.8860% +5.5658%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
send_zst/10             time:   [87.412 us 88.711 us 90.077 us]
                        change: [+4.7130% +6.8866% +9.1067%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 7 outliers among 100 measurements (7.00%)
  6 (6.00%) high mild
  1 (1.00%) high severe
send_zst/100            time:   [913.25 us 925.94 us 939.17 us]
                        change: [+6.1163% +7.5650% +9.1937%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
send_zst/1000           time:   [8.9915 ms 9.1422 ms 9.3021 ms]
                        change: [+5.1674% +7.0818% +8.9474%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 14 outliers among 100 measurements (14.00%)
  14 (14.00%) high mild
```

So there seems to be a performance regression of about 3-7% but I am not quite sure where it is coming from.

thomaseizinger:
See https://github.com/Restioson/xtra/pull/52 for how I benchmarked.

Restioson:
> I've had a look and I am not sure if we are using criterion correctly there? The setup_fn is not benchmarked if I understand correctly, only the final Finish message is. Is that on purpose?

As far as this goes, I just wanted to benchmark the time to process the messages. Come to think of it, this is _not_ how `crude_bench.rs` currently works though - I'm not actually sure which is more useful of a measure. The main difference is that `crude_bench` lets the actor begin processing while messages are being sent, whereas this does not. This makes sense though as to why there is a difference between `crude_bench` and criterion. It might be worth including multiple benchmarks that measure that (of course with criterion we can have as many as we like!) With the current criterion benchmark I wanted mainly to bench how fast it'd be to receive and process 100 messages, i.e the management loop overhead of doing that receive with 0 contention, since that's all I'd expect to change with these changes.

Restioson:
> Are you squash merging in this repo

Yes, I am squash merging

thomaseizinger:
> With the current criterion benchmark I wanted mainly to bench how fast it'd be to receive and process 100 messages, i.e the management loop overhead of doing that receive with 0 contention, since that's all I'd expect to change with these changes.

I guess my benchmark from above still achieves this? Regardless of how exactly we are benchmarking, with the one I did above, this branch is 7% slower to process 1000 messages compared to the previous version. Is that acceptable?

thomaseizinger:
I've pushed some more changes. The last commit (https://github.com/Restioson/xtra/pull/51/commits/894e440daf473823f2814a1c6871537258c57bb8) might be controversial :D

Let me know what you think.

thomaseizinger:
Do you think https://github.com/thomaseizinger/xtra/blob/c63eac080b5c4c488d25b193486ce9b740f91b7b/src/context.rs#L215-L219 is still needed? The next that happens right after this is that we are waiting for a message on the inbox and if it is a `Shutdown` message, we would immediately exit.

I guess you can see what I am getting at. If we can trim this `run` loop down a bit further, we might be able to expose its primitives (`tick` and `inbox`) publicly and then anyone can build themselves a custom event loop!

thomaseizinger:
Also, what do you think of getting rid of https://github.com/thomaseizinger/xtra/blob/c63eac080b5c4c488d25b193486ce9b740f91b7b/src/context.rs#L208-L213 by moving the check into `inbox` and returning an empty stream in case the actor is shutdown?

Then the `while let Some` loop would immediately exit and we would call `actor#stopped`, same as we are currently doing.

Should probably add a test for that piece of logic :)

src/context.rs
```diff
@@ -468,6 +398,32 @@ impl<A: Actor> Context<A> {
...
+            futures_util::stream::unfold(broadcast_rx, |broadcast_rx| async move {
+                match broadcast_rx.recv_async().await {
+                    Ok(msg) => Some((msg, broadcast_rx)),
+                    Err(_) => None,
+                }
+            })
+            .map(Either::Left)
+            .boxed();
+
+        let address_stream = addr_rx.into_stream().map(Either::Right).boxed();
```
Restioson, line 416: This is unfortunately a very bad idea since streams have very unfavourable performance characteristics; I can elaborate on it a bit later

Restioson: Ok, here is the problem with streams as I understand it. Most of this comes with a conversation I had with zesterer (the main developer of flume) and stjepang about streams and channels.

With a future-based receive, you as the channel writer have a way to *roughly*[^1] de-register interest in receiving. How channels such as flume works is that when you call `recv`, if there isn't a message, it will put the waker into a queue of receivers interested in receiving a message. Then, when a message is sent, one is popped off, the message is delivered to it, and the waker is woken. When a future is dropped, their waker is removed from the queue, regardless of if it has received a message. So, to summarise, exactly one receiver is woken to take exactly one message.

However, with a stream, it is added to a separate queue, and streams are woken for *every message*, since there is no way to know if a stream is still interested, as you don't control the stream's receive future implementation. This creates an unavoidable thundering herd problem. I'm not aware of any way to work around this - flume has this property, as does async-channel.

[^1]: the Drop implementation of the receive future is usually used to see when a receiver is de-registering interest. You *can* `mem::forget` a future but the general consensus is that if you do this you void your guarantees on receiving all messages.

Restioson:
> Also, what do you think of getting rid of https://github.com/thomaseizinger/xtra/blob/c63eac080b5c4c488d25b193486ce9b740f91b7b/src/context.rs#L208-L213 by moving the check into inbox and returning an empty stream in case the actor is shutdown?

That seems okay to me. 

> Then run could really just look like this 😍 :

Awesome!

src/context.rs
```diff
@@ -468,6 +398,32 @@ impl<A: Actor> Context<A> {
 
         Ok(fut)
     }
+
+    fn inbox(&self) -> impl FusedStream<Item = Either<BroadcastMessage<A>, AddressMessage<A>>> {
```
Restioson, line 402: I wonder if BroadcastMessage and AddressMessage could be merged together, having a third variant for BroadcastMessageEnvelope. Could check if that changes the size (it might due to niches, I'm not sure).

Restioson: Also, if this is going to be public, it might be worth considering wrapping the item in an opaque struct so that the loop can't see inside. However, this might be a good feature (telling between Shutdown and normal messages). On the other hand, it might open up ways for people to somewhat misuse the rest of the lifecycle infrastructure; the goal should be to create a pit of success, and having one and only one way to manage the lifecycle is probably a better way to do that

src/context.rs
thomaseizinger: Thanks for the thorough explanation!

If I understand correctly, we should be fine though if we base both streams on top of `unfold` and the `async_recv` function? That way, only the immediate next item is registered and if we drop the stream, the future will be dropped as well.

thomaseizinger: Also, would you be open to the idea of completely encapsulating this API so the streams are an implementation detail and merge something like this despite non-ideal performance characteristics?

Once they are an implementation detail, we can iterate on internal designs to improve performance and have the API out there.

Restioson: Maybe, it depends exactly how its done. Would you mind writing it out or pushing what you have in mind exactly?

thomaseizinger: > Maybe, it depends exactly how its done. Would you mind writing it out or pushing what you have in mind exactly?

See https://github.com/Restioson/xtra/pull/51/commits/d76c2f82132a25df7c76d411681ae7de59df1d21.

src/context.rs
```diff
@@ -399,26 +402,59 @@ impl<A: Actor> Context<A> {
...
-    fn inbox(&self) -> impl FusedStream<Item = Either<BroadcastMessage<A>, AddressMessage<A>>> {
-        let addr_rx = self.receiver.clone();
-        let broadcast_rx = self.broadcast_receiver.clone();
+    fn inbox(&self) -> Inbox<A> {
+        Inbox::new(self.receiver.clone(), self.broadcast_receiver.clone())
+    }
+}
+
+struct Inbox<A> {
+    inner: BoxStream<'static, Either<BroadcastMessage<A>, AddressMessage<A>>>,
```
Restioson, line 411: small nit: it might be less overhead and allocation to not use a box and just store what we need inline and then manually implement stream - can see how the benches look

thomaseizinger: Would you mind if we optimize these things later?

src/context.rs
```diff
@@ -206,7 +208,6 @@ impl<A: Actor> Context<A> {
         // Idk why anyone would do this, but we have to check that they didn't do ctx.stop()
         // in the started method, otherwise it would kinda be a bug
         if !self.check_running(&mut actor).await {
-            self.stop_all();
             actor.stopped().await;
             return;
         }
```
thomaseizinger, line 213: I think this should logically go into the `tick` function because that is where this fn is already called and that one already takes `&mut A`. I tried doing that by emitting a special `InboxMessage` if we instantiate `Inbox` while `RunningState::Stopping` but it gets a bit messy.

We do have a test for it now so it should be safe to iterate on some designs!

src/context.rs
```diff
@@ -217,71 +218,31 @@ impl<A: Actor> Context<A> {
...
 
     /// Handle a message and immediate notifications, returning whether to exit from the manage loop
     /// or not.
-    async fn tick(
-        &mut self,
-        msg: Either<BroadcastMessage<A>, AddressMessage<A>>,
-        actor: &mut A,
-    ) -> ContinueManageLoop {
-        match msg {
+    async fn tick(&mut self, msg: InboxMessage<A>, actor: &mut A) -> ContinueManageLoop {
```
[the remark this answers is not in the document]
Restioson: Yep, `ControlFlow` could definitely be used here. I think this could become public.

src/context.rs
Restioson: Yea, if the peformance isn't too bad then we can leave it for now

src/context.rs
```diff
@@ -468,6 +401,65 @@ impl<A: Actor> Context<A> {
...
+
+    pub async fn next(&mut self) -> Option<InboxMessage<A>> {
+        let inner = self.inner.next().await?;
+
+        Some(InboxMessage { inner })
+    }
+}
+
+struct InboxMessage<A> {
+    inner: Either<BroadcastMessage<A>, AddressMessage<A>>,
```
Restioson, line 455: I wonder if it'd be simpler to just merge BroadcastMessage<A> and AddressMessage<A> as such:

```rust
enum BroadcastMessage<A: Actor> {
    Message(Box<dyn MessageEnvelope<Actor = A>>),
    Broadcast(Box<dyn BroadcastMessageEnvelope<Actor = A>>),
    Shutdown,
    LastAddress,
}
```

thomaseizinger: Yeah that sounds like a good idea to me! I'd still wrap it in `InboxMessage` though so that enum details remain an implementation detail of the library.

src/context.rs
Restioson: Sounds good to me

src/context.rs
thomaseizinger: > I wonder if it'd be simpler to just merge BroadcastMessage and AddressMessage as such:
> 
> ```rust
> enum Message<A: Actor> {
>     Message(Box<dyn MessageEnvelope<Actor = A>>),
>     Broadcast(Box<dyn BroadcastMessageEnvelope<Actor = A>>),
>     Shutdown,
>     LastAddress,
> }
> ```

Unfortunately it is not that easy because `barrage::Sender` for example enforces that `T` is `Clone` and `MessageEnvelope` cannot be cloned. We could combine it _just_ for whatever is in `InboxMessage` but I don' think it is worth it.

thomaseizinger:
Are you attached to `handle_while` staying? With `inbox` and `tick` being public, it is at possible for other people to write their own event loop and they could fairly easily write themselves a loop with `futures::select` that achieves the same thing as `handle_while`.

thomaseizinger:
Argh, not quite there. Had to re-add `check_running` for now to fix a bug highlighted by the new test.

src/context.rs
Restioson: Ah, right. Unfortunate, we can't really relax the Clone requirement either as it is not possible conceptually to have broadcast without Cloning (or an ugly solution involving `(FnOnce() -> T)` + Clone closures)

Restioson:
> Are you attached to handle_while staying? With inbox and tick being public, it is at possible for other people to write their own event loop and they could fairly easily write themselves a loop with futures::select that achieves the same thing as handle_while.

Is there any disadvantage to keeping handle_while except for needing to keep some code there? There is a slight difference in that handle_while can be inside only one message handler, and can use the message itself for info or to construct the future

thomaseizinger:
> There is a slight difference in that handle_while can be inside only one message handler, and can use the message itself for info or to construct the future

I had a look at the example that is linked for `handle_while` and was honestly confused on what it is meant to achieve. I applied the following patch to the example and the behaviour is idential:

```patch
Index: examples/interleaved_messages.rs
<+>UTF-8
===================================================================
diff --git a/examples/interleaved_messages.rs b/examples/interleaved_messages.rs
--- a/examples/interleaved_messages.rs	(revision 167900fd4a6dc58f3af544a686ee2e14d542bdb8)
+++ b/examples/interleaved_messages.rs	(date 1640298539198)
@@ -20,9 +20,7 @@
 impl Handler<Hello> for ActorA {
     async fn handle(&mut self, _: Hello, ctx: &mut Context<Self>) {
         println!("ActorA: Hello");
-        ctx.handle_while(self, self.actor_b.send(Hello))
-            .await
-            .unwrap();
+        self.actor_b.send(Hello).await;
     }
 }
```

In the example, `ActorA` is already [spawned](https://github.com/Restioson/xtra/blob/167900fd4a6dc58f3af544a686ee2e14d542bdb8/examples/interleaved_messages.rs#L51-L55) into a global runtime, meaning there is already an event loop running that will make it receive messages. While the `Hello` handler is active, it cannot receive any more messages (that would be unsound because we would have two active `&mut`). That handler is not going to complete until the `handle_while` future returns.

From what I understand, `handle_while` is an _alternative_ to calling `run` on `Context`. You can't have two event-loops running for the same actor so using it from within an `Actor` seems wrong?

Restioson:
I'll take a look at this in a second but the original motivation was #13

thomaseizinger:
> I'll take a look at this in a second but the original motivation was #13

Thanks for linking this! That is actually exactly what I was getting at with https://github.com/Restioson/xtra/issues/49. We've solved these issues by having one of the actors use `do_send_async` instead of `send`. That way, the handler can complete and the deadlock is resolved.

thomaseizinger:
In response to https://github.com/Restioson/xtra/issues/13#issuecomment-641363353.

> Unfortunately it does not - I do need to use `send` and await it since the return values are required.

This could be solved by having `do_send` / `do_send_async` return a oneshot channel that can again be awaited for the response but have the handler execute asynchronously in the background.

Restioson:
Yea, that's essentially what handle while does. It has the handler execute other messages while it's waiting for a given future (the one returned by send in this case) to complete. I don't quite understand what you mean by this I think

thomaseizinger:
> Yea, that's essentially what handle while does. It has the handler execute other messages while it's waiting for a given future (the one returned by send in this case) to complete. I don't quite understand what you mean by this I think

I finally understood how this works, sorry about all of these weirdness! I now fully get the usecase of `handle_while` and have to say it is quite neat! On the other hand though, I do think it can also be a foot-gun. Here is my thinking:

1. Actors are meant to be a safe concurrency pattern
2. At least I came into `xtra` with the assumption that, this being Rust with all its safe concurrency and no shared mutable state features, handlers of actors are atomic in regards to state changes of an actor. In other words, a handler can assume nothing messes with its state from top to bottom of the handler.

Once you start using `handle_while`, (2) is no longer true and almost becomes dangerous to use if you've previously relied on this invariant and someone comes in and introduces a `handle_while`. What do you think about that? Personally, I'd likely avoid the use of `handle_while` because of that and instead try and change the design to rely on async message delivery.

Restioson:
As far as 2 goes, I'm okay with breaking that invariant! since it's explicit, and the use case is uncommon. Somewhat like how RefCell and atomics break the "no modification behind &" invariant, but that's okay because it's explicit and uncommon. It was mainly introduced to allow people to solve circular chains of messages like in #13. How else could this be achieved do you think?

thomaseizinger:
> How else could this be achieved do you think?

I can't think of a way of how it can be solved differently.

I am going to cleanup the remainder of this PR soon so we can push it over the finish line :)

Restioson:
Thanks! A followup PR could always address it as required. I'll be back in the city by the end of next week so I can review it then.

thomaseizinger:
There is some stuff in here that I would like to get merged but I am not really happy with the fact that we can't remove the check to `self.check_running`. Without making that public, people can't actually implement their own event loop which means we are just asking for buggy code down the line here.

I'll see if I can think of a way of fixing this. Might need to re-work a bit, how the whole lifecycle of an actor is tracked.

src/context.rs
```diff
@@ -203,107 +206,67 @@ impl<A: Actor> Context<A> {
...
-            self.stop_all();
             return actor.stopped().await;
         }
 
-        // Similar to above
-        if let Some(BroadcastMessage::Shutdown) = self.broadcast_receiver.try_recv().unwrap() {
-            return actor.stopped().await;
-        }
+        // // Similar to above
+        // if let Some(BroadcastMessage::Shutdown) = self.broadcast_receiver.try_recv().unwrap() {
```
Restioson, line 214: I've just realised that there is a bug on master relating to this - it will swallow a `Message` variant from other actors on the same address, so that extra case should probably be handled. Regardless of that, I don't think this should go, in case of this actor being attached to the context late. Imagine:

Actor A starts
Actor A spawns actor B and spawns it onto an executor
Actor A broadcasts shutdown
Actor B's task is polled for the first time, and if this code isn't here, it will not shut down properly

src/context.rs
```diff
@@ -467,6 +416,60 @@ impl<A: Actor> Context<A> {
...
+                match address_receiver.recv_async().await {
+                    Ok(msg) => Some((msg, address_receiver)),
+                    Err(_) => None,
+                }
+            })
+            .map(Either::Right)
+            .boxed();
+
+        Self {
+            inner: futures_util::stream_select!(broadcast_stream, address_stream).boxed(),
```
Restioson, line 458: The fact that stream_select is pseudorandom could cause potential issues with broadcast shutdown messages - if there is an external address constantly sending messages, the broadcast stream could theoretically be starved.

thomaseizinger:
Going to close this because too many things are changing in #85.

PR #52, opened 2021-12-20
Use criterion's async support

thomaseizinger:


benches/throughput.rs
```diff
@@ -30,25 +31,20 @@ impl Handler<Finish> for Counter {
...
+    let mut group = c.benchmark_group("send_zst");
+
+    for num_messages in [1, 10, 100, 1000] {
+        let (address, task) = Counter(0).create(Some(num_messages)).run();
+        let _task = smol::spawn(task);
+
+        group.bench_with_input(BenchmarkId::from_parameter(num_messages), &num_messages, |b, &num_messages| {
+            b.to_async(SmolExecutor).iter(|| async {
+                for _ in 0..num_messages {
+                    address.send(IncrementZst).await;
```
Restioson, line 43: I think this will have much more overhead included, as after every message the future yields, and then returns.

Restioson: Looking at the other PR comment now:

> send_zst/1              time:   [8.5991 us 8.7618 us 8.9253 us]

This would imply a time of processing of 8us per message, which seems very high compared to what it was being measured at before at ~210ns per message average. Some of this may be due to the channel sending, but I think that order of magnitude difference might be too much for that... Maybe a `do_send` benchmark could be included too?

thomaseizinger: > It depends how this actually benchmarks, though, and I'm not familiar with criterion - what does it measure in this?

Are the details important as long as we are always benchmarking in the same way? This way felt natural to me if we want to benchmark message processing.

thomaseizinger: > I think this will have much more overhead included, as after every message the future yields, and then returns. It depends how this actually benchmarks, though, and I'm not familiar with criterion - what does it measure in this? Maybe it measures time-from-last-poll? Maybe it already accounts for overhead? In that case, it might be okay.

I don't know the answer to these questions but the performance here might also be runtime dependent.

My naive viewpoint would be that this is a benchmark closer to the real world as it actually dispatches the message via `.await`, meaning an application will take ~8us to process sending a message with `xtra` as its actor library.

Should we compare how `xtra` performs across popular runtimes?

The annoying thing about measuring performance this way is that our numbers will be impacted by the runtime improving performance but at the same time, we get results that are closer to the real world. Not sure whether that is positive overall!

Restioson: > My naive viewpoint would be that this is a benchmark closer to the real world as it actually dispatches the message via .await, meaning an application will take ~8us to process sending a message with xtra as its actor library.

This is true. When I originally wrote `crude_bench.rs` I set out to essentially measure how much overhead xtra would contribute, not necessarily how much time it would take to process a single message and wait for a reply, since to measure that is often just a runtime benchmark. 

> The annoying thing about measuring performance this way is that our numbers will be impacted by the runtime improving performance but at the same time, we get results that are closer to the real world. Not sure whether that is positive overall!

That's true. 

At least if we have a baseline of comparing the same way it should be somewhat consistent. I'd just be concerned that it would introduce noise to the benchmark.

Restioson:
If you could format so it passes CI this I'd be ok with merging

thomaseizinger:
> If you could format so it passes CI this I'd be ok with merging

Done!

Restioson:
Okay, this is still failing CI, but it's on an unrelated part of the code and I can't reproduce this locally. Also, the same test runs twice, but one time it fails, and one time it passes. I think this is probably spurious?

Restioson:
Going to go ahead and merge this. We can always experiment and add more benches later if it matters.

PR #55, opened 2022-02-03
Move `Actor::create` to dedicated trait

thomaseizinger:
This function is not meant to ever be implemented by users.
This is indicated by all the fields in `ActorManager` being
`pub(crate)`. To avoid this confusion, we can move this
functionality to a dedicated trait that lives within the
`manager` module which allows us to make the fields private
altogeher.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,7 +31,7 @@ pub mod prelude {
     #[cfg(feature = "with-tracing-0_1")]
     pub use crate::tracing::InstrumentedExt;
     #[doc(no_inline)]
-    pub use crate::{Actor, Handler, Message};
+    pub use crate::{manager::Create, Actor, Handler, Message};
 }
 
 /// A message that can be sent to an [`Actor`](trait.Actor.html) for processing. They are processed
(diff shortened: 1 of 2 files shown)


Restioson:
This could make sense, but I'm not sure it's worth the ergonomic change in having to import `Create` everywhere. As far as I can tell, some stdlib traits also have functions like this which are never intended to be implemented by the user - see [`Iterator::chain`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.chain).

thomaseizinger:
> This could make sense, but I'm not sure it's worth the ergonomic change in having to import `Create` everywhere. As far as I can tell, some stdlib traits also have functions like this which are never intended to be implemented by the user - see [`Iterator::chain`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.chain).

Up to you. If people use `xtra;:prelude::*`, there is no change needed as seen by the examples that all still work!

thomaseizinger:
> This could make sense, but I'm not sure it's worth the ergonomic change in having to import Create everywhere.

In my experience, code that instantiates actors is usually in a different module than the one defining them.

Defining an actor obviously needs to import `Actor`. Code that creates them currently also has to import `Actor` to call the `create` method.

I usually end up doing `use xtra::Actor as _;` to not pollute the namespace. IMO it would make more sense if one could always import `Create` that way (`use xtra::Create as _;`) and `Actor` only if you are defining one.

PR #56, opened 2022-02-03
Fix dead code warnings

thomaseizinger:
In case no features are enabled, these 3 imports
are flagged as unused code.

Add the import line to each individual module to
avoid this.


thomaseizinger:
I don't understand why CI is failing on this :(

Restioson:
Passing now so will merge it.

PR #57, opened 2022-02-04
Re-export `async_trait` from prelude

thomaseizinger:
This allows downstream users to avoid declaring this dependency
separately.

### DIFF
--- a/examples/basic_async_std.rs
+++ b/examples/basic_async_std.rs
@@ -18,7 +18,7 @@ impl Message for Print {
     type Result = ();
 }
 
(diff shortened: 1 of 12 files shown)


Restioson:
Seems like CI is failing here on an unrelated part of the code? Weird...

thomaseizinger:
> Seems like CI is failing here on an unrelated part of the code? Weird...

Yes, this test seems to be a bit unstable. I've had it fail on several of my PRs today.

Restioson:
Happy to merge this for now then

PR #59, opened 2022-02-04
Introduce `Actor::Stop`

thomaseizinger:
This is useful for building supervisor patterns where the supervisor can learn, why a certain actor stopped operating. You can find our current implementation here: https://github.com/itchysats/itchysats/blob/feature/better-supervisor/xtras/src/supervisor.rs

A couple of design decisions went into this patch:

1. I chose to be Rust-stable compatible and thus, we don't have associated type-defaults. Once they become stable, we could use them.
2. I chose to require `Actor::stopped` to be implemented by every actor. This is a slight ergonomic hit but I believe it is the overall best trade-off.
a. A developer SHOULD think about why its actor is stopping.
b. If `Stop` is `()`, the method body can be empty and usually collapses into one line, hence not that big a deal really.
c. If `Stop` ever changes, the compiler will complain and lead the developer towards thinking again about what should happen in `stopped`.

We could require `Stop` to implement `Default`. I played with this idea but I think it is a foot-gun. If an actor implementation chooses to set a `Stop` value, I believe the compiler should complain at every usage that it needs to be handled. Forcing `Stop` to implement `Default` would allow things to quietly work and the value would likely get dropped somewhere.

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -48,14 +48,16 @@ pub struct ActorManager<A: Actor> {
     pub(crate) ctx: Context<A>,
 }
 
-impl<A: Actor> ActorManager<A> {
+impl<A: Actor<Stop = ()>> ActorManager<A> {
     /// Spawn the actor's main loop on the given runtime. This will allow it to handle messages.
     pub fn spawn<S: Spawner>(self, spawner: &mut S) -> Address<A> {
         let (addr, fut) = self.run();
         spawner.spawn(fut);
         addr
     }
+}
 
+impl<A: Actor> ActorManager<A> {
     /// Starts the manager loop, returning the actor's address and its manage future. This will
     /// start the actor and allow it to respond to messages.
     ///
@@ -64,14 +66,14 @@ impl<A: Actor> ActorManager<A> {
     /// ```no_run
     /// # use xtra::prelude::*;
     /// struct MyActor;
-    /// impl Actor for MyActor {}
+    /// # #[async_trait::async_trait] impl Actor for MyActor {type Stop = (); async fn stopped(self) -> Self::Stop {} }
     ///
     /// smol::block_on(async {
     ///     let (addr, fut) = MyActor.create(None).run();
     ///     smol::spawn(fut).detach(); // Actually spawn the actor onto an executor
     /// });
     /// ```
-    pub fn run(self) -> (Address<A>, impl Future<Output = ()>) {
+    pub fn run(self) -> (Address<A>, impl Future<Output = A::Stop>) {
         (self.address, self.ctx.run(self.actor))
     }
 }
--- a/examples/basic_smol.rs
+++ b/examples/basic_smol.rs
@@ -15,7 +15,7 @@ impl Printer {
 impl Actor for Printer {
     type Stop = ();
 
-    async fn stopped(self) -> Self::Stop { }
+    async fn stopped(self) -> Self::Stop {}
 }
 
 struct Print(String);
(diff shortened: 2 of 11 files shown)


src/manager.rs
```diff
@@ -48,14 +48,16 @@ pub struct ActorManager<A: Actor> {
     pub(crate) ctx: Context<A>,
 }
 
-impl<A: Actor> ActorManager<A> {
+impl<A: Actor<Stop = ()>> ActorManager<A> {
```
Restioson, line 51: Should this perhaps not accept any stop type?

Restioson:
Other than the comment I left and that CI is failing (on rustfmt), LGTM. Anything else to add here or ready for merge soon?

src/manager.rs
thomaseizinger: I've split it out because we would have to do:

```rust
spawner.spawn(async move {
	let _ = fut.await;
});
```

I am not sure if it is good to drop the returned value here?

thomaseizinger:
> Other than the comment I left and that CI is failing (on rustfmt), LGTM. Anything else to add here or ready for merge soon?

Sorry! Must have messed up the cherry-picking. I am okay with letting this sit for a while if you want. I've literally only extended our supervisor implementation today to use this and it is not yet merged. So if you want, we can merge this first to gain some more experience with this functionality.

src/manager.rs
Restioson: Hm. I think there's something to be said for having a function that takes an actor and returns an address - I think that's probably a fairly common usecase. Thus far anyway, in all the code I've written I haven't needed to return anything from `stopped`, so I think that goes to show that there are usecases where you don't care about the return value. If you *do* care about it and still want to `spawn`, one could do:

```rust
let (addr, fut) = actor.create(/* cap */).run();
let ret = tokio::spawn(fut);
```

Otherwise, there could be yet another method like
```rust
let (addr, ret) = actor.create(/* cap */).foobar();
```

which does this for you. I'm not sure if this is needed though? This would require some internal oneshot channel type thing to abstract over all the different task abstractions. It'd also be different to a `smol` `Task` for instance, since dropping it would not cancel it, which might be weird. Thoughts?

thomaseizinger: > I think there's something to be said for having a function that takes an actor and returns an address

Definitely. And as long as `Stop` is set to `()`, nothing changes with this patch. You can still use `.spawn()`.

I don't think it is worth creating another method for spawning and returning. For the uncommon case where `Stop` returns something other than `()`, the additional `.run()` is barely worth mentioning the complexity IMO.

On the other hand, if we make this function take actors with any `Stop` value, we are removing an ability for the compiler to guide the user towards what is likely the correct behaviour: Caring about the return value if it is `!= ()`.

thomaseizinger: > I think there's something to be said for having a function that takes an actor and returns an address

Definitely. And as long as `Stop` is set to `()`, nothing changes with this patch. You can still use `.spawn()`.

I don't think it is worth creating another method for spawning and returning. For the uncommon case where `Stop` returns something other than `()`, the additional `.run()` is barely worth mentioning the complexity IMO.

On the other hand, if we make this function take actors with any `Stop` value, we are removing an ability for the compiler to guide the user towards what is likely the correct behaviour: Caring about the return value if it is `!= ()`.

Restioson: > On the other hand, if we make this function take actors with any Stop value, we are removing an ability for the compiler to guide the user towards what is likely the correct behaviour: Caring about the return value if it is != ().

That's true. We can keep it this way then I think.

Restioson:
It's a fairly simple change on the grand scale of things so I think we can merge it now. It can always be iterated on in later pull requests.

PR #60, opened 2022-02-22
Fix `is_connected` for an `Actor` returning `StopSelf` in `stopping`

thomaseizinger:


Restioson:
Ah, I see what you mean by this issue. If you can get it fixed I'll merge it, but unfortunately right now my bandwidth to write code myself is a little low due to university term and life 😅

thomaseizinger:
> Ah, I see what you mean by this issue. If you can get it fixed I'll merge it, but unfortunately right now my bandwidth to write code myself is a little low due to university term and life 😅 

I am happy to write the code! Do you have an idea of how you want it fixed? It is not obvious when I looked at it :)

Is there a way of figuring out how many actors have been created on an address? Should we introduce a AtomicU32 that is shared across all contexts? I don't think we can use the address because that one could be cloned additionally!

Restioson:
That could work, however, I wonder if something from the channels could be used? For instance, checking if dropping the address leads to the channel disconnecting. This could be either the broadcast or flume channel.

thomaseizinger:
That sounds like a good idea. I'll explore that!

thomaseizinger:
I've changed the implementation of `Address::is_connected` to use the `Sender` instead of the `RefCounter`. This fixes the test I added. What is your confidence on the remaining test suite that this works as expected in all circumstances?

thomaseizinger:
Friendly ping @Restioson :)

Restioson:
Oh sorry, I missed this. will take a look

thomaseizinger:
> Oh sorry, I missed this. will take a look. Seems to still fail on smol though?

That is our famous flaky test, unrelated to this code change!

Restioson:
> That is our famous flaky test, unrelated to this code change!

Seems like two failed?

Restioson:
Rerunning it now

thomaseizinger:
Now it seems to be hanging forever. Odd, they all passed on my machine ...

thomaseizinger:
Force-pushed to re-trigger CI.

Restioson:
Smol still seems to hang but everything else works

thomaseizinger:
I can't reproduce this locally :(

Will have a deeper look at the code later but this test has failed on me in CI a couple of times!

Restioson:
Interesting it only fails with smol 🤔

thomaseizinger:
> Interesting it only fails with smol thinking

These tests are not run with any other runtime AFAIK.

thomaseizinger:
> > Interesting it only fails with smol thinking
> 
> These tests are not run with any other runtime AFAIK. At least they use `smol_potat` which I assume only works with smol!

Yep, tokio runs `0` tests for example: https://github.com/Restioson/xtra/runs/5476184559?check_suite_focus=true#step:7:70

Restioson:
I see, that makes sense then. I forgot that the other CI pipelines are just there to make sure it builds, since it's assumed that if it works on one executor, it should work on all of them

Restioson:
I wonder if perhaps something about `#[smol_potat::test]` is making this fail. Might be worth trying to initialise the executor manually. Apologies by the way that I can't work much on this right now, it's just that university term is in full swing 😓

thomaseizinger:
I'll see to add some tests using tokio and see if that passes.

thomaseizinger:
@Restioson It seems everything passes with the tokio executor! What do you want to do about this situation? :D

Restioson:
This seems to be some kind of smol scheduling issue, in that case. Happy to merge!

PR #62, opened 2022-04-09
Remove `Message` trait

thomaseizinger:
We have been working with `xtra` for a while now and something that came up repeatedly is the `Message` trait. In particular, I would like to discuss the idea of removing it.

Requiring the trait to be implemented in order to declare that an actor is able to handle a message hurts ergonomics as you try to modularise a codebase that heavily depends on `xtra`. Traits can only be implemented once and the orphan rule enforces that this happens where the struct is declared.

This PR presents a Proof-of-Concept for removing the `Message` trait in favor of an associated type on the `Handler` trait. Unless I missed something, all current functionality is retained. What do you think?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -34,6 +34,8 @@ pub mod prelude {
     pub use crate::tracing::InstrumentedExt;
     #[doc(no_inline)]
     pub use crate::{Actor, Handler};
+
+    pub use async_trait::async_trait;
 }
 
 /// A trait indicating that an [`Actor`](trait.Actor.html) can handle a given [`Message`](trait.Message.html)
@@ -69,7 +71,7 @@ pub mod prelude {
 ///     })
 /// }
 /// ```
-#[async_trait::async_trait]
+#[async_trait]
 pub trait Handler<M>: Actor {
     /// The return value of this handler.
     type Return: Send + 'static;
@@ -140,7 +142,7 @@ pub trait Handler<M>: Actor {
 /// ```
 ///
 /// For longer examples, see the `examples` directory.
-#[async_trait::async_trait]
+#[async_trait]
 pub trait Actor: 'static + Send + Sized {
     /// Value returned from the actor when [`Actor::stopped`] is called.
     type Stop: Send + 'static;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -71,7 +71,7 @@ pub mod prelude {
 ///     })
 /// }
 /// ```
-#[async_trait]
+#[async_trait::async_trait]
 pub trait Handler<M>: Actor {
     /// The return value of this handler.
     type Return: Send + 'static;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -142,7 +142,7 @@ pub trait Handler<M>: Actor {
 /// ```
 ///
 /// For longer examples, see the `examples` directory.
-#[async_trait]
+#[async_trait::async_trait]
 pub trait Actor: 'static + Send + Sized {
     /// Value returned from the actor when [`Actor::stopped`] is called.
     type Stop: Send + 'static;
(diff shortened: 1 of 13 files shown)


Restioson:
I think overall, I like this change. I'll take a closer look at the code itself a bit later. I like how this should also reduce the need for newtypes of messages to allow them to be handled (I've done this a lot for web socket packet frame messages).

thomaseizinger:
> I think overall, I like this change. It improves ergonomics and simplifies the code, and the arguments you've presented make a lot of sense. I'll take a closer look at the code itself a bit later. I like how this should also reduce the need for newtypes of messages to allow them to be handled (I've done this a lot for web socket packet frame messages).

Cool, that is great to hear! Looking forward to any feedback you might have :)

src/message_channel.rs
```diff
@@ -101,7 +101,7 @@ impl<M: Message> Future for SendFuture<M> {
 ///     })
 /// }
 /// ```
-pub trait MessageChannel<M: Message>: Sealed + Unpin + Send + Sync {
+pub trait MessageChannel<M, R>: Sealed + Unpin + Send + Sync {
```
thomaseizinger, line 104: Perhaps `R` can be an associated type here as well. I'll try!

thomaseizinger:
@Restioson I've pushed another commit that also uses an associated type for the `MessageChannel`. I think that is more correct because `Handler` also uses an associated type. Now here is a question though:

Is an associated type the way to go? At the moment, we are effectively saying: An actor can only have one handler per message type and the return type is given based on that.

We could also allow multiple handlers for a type with a different return type each. However, now that I think about it, that would likely cause weird ergonomics because it forces the _caller_ of `Address#send` to be explicit about the return type and _based_ on the return type, we might dispatch to a different handler. That doesn't sound good haha.

Restioson:
> Is an associated type the way to go? At the moment, we are effectively saying: An actor can only have one handler per message type and the return type is given based on that.

This is the status quo, as with `Handler<M>` before, and the associated type of `M::Result`.

> We could also allow multiple handlers for a type with a different return type each. However, now that I think about it, that would likely cause weird ergonomics because it forces the caller of Address#send to be explicit about the return type and based on the return type, we might dispatch to a different handler. That doesn't sound good haha.

I agree. I think this is one of the reasons that an associated type was chosen for `Iterator` - else, `next` and friends would require a lot of disambiguation. I think if a different handler wants to handle a message, a newtype could be used, and that would require less turbofish and be less confusing - one handler to one message type seems intuitive.

thomaseizinger:
Thanks for confirming!

With that, this PR is ready for review ☺️

src/lib.rs
```diff
@@ -91,12 +70,15 @@ pub trait Message: Send + 'static {
 /// }
 /// ```
 #[async_trait::async_trait]
-pub trait Handler<M: Message>: Actor {
+pub trait Handler<M>: Actor {
+    /// The return value of this handler.
+    type Return: Send + 'static;
```
Restioson, line 75: Should this be kept as Result to keep it with the old naming convention?

thomaseizinger: Sure, I don't mind :)

I figured `Return` is slightly more appropriate because `Result` in Rust usually refers to the enum and we can have `Handler`s that do _not_ return a `Result` but just `()` or are otherwise infallible.

I don't feel strongly about that though!

Restioson: Fair enough. I borrowed Result from Actix I think. Likewise though, I don't feel very strongly about it, so we can keep it as Return as far as I'm concerned

thomaseizinger: If I get the choice I'll stick with `Return` :)

Restioson: I think that makes sense

Restioson (approved):

PR #65, opened 2022-04-10
Use criterion for benchmarks

Restioson:
Add some benchmarks using criterion (thanks to @thomaseizinger for #52)


Restioson:
Oh no, I'd forgotten what was on my local branch - oops

PR #66, opened 2022-04-10
Use criterion for benchmarks (correct git history)

Restioson:
#65 but cherrypicked properly

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,8 @@ smol = "1.1"
 smol-potat = "1.1"
 smol-timeout = "0.6"
 waker-fn = "1.1"
-criterion = { version = "0.3", features = ["async_smol"] }
+criterion = { version = "0.3", features = ["async_tokio"] }
+tokio = { version = "1.0", features = ["macros", "rt-multi-thread"] }
 
 [features]
 default = ["timing"]
@@ -89,7 +90,7 @@ required-features = ["with-smol-1"]
 
 [[test]]
 name = "basic"
-required-features = ["with-smol-1"]
+required-features = ["with-tokio-1"]
 
 [workspace]
 members = ["examples/basic_wasm_bindgen"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ smol = "1.1"
 smol-potat = "1.1"
 smol-timeout = "0.6"
 waker-fn = "1.1"
-tokio = { version = "1.0", features = ["macros"] }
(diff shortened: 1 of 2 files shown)


Restioson:
Tried to cherrypick this from the old criterion branch but I seem to have messed up the signing in the process. Is that okay @thomaseizinger ?

Restioson:
Also, I'm interested as to whether you think we get more relevant numbers from [0166717](https://github.com/Restioson/xtra/pull/66/commits/0166717d70d107995318c8d6367bbe4590c671cc). The pattern I was seeing before was just that each n would just be 10x higher than the last, which is exactly as to be expected, so it was not very useful to have the multiple levels without batching like so.

thomaseizinger:
> Tried to cherrypick this from the old criterion branch but I seem to have messed up the signing in the process. Is that okay @thomaseizinger ? 

Yeah that is totally fine :)
Cherry-picking creates a new commit so it is natural that the signature is broken afterwards.

thomaseizinger:
> Also, I'm interested as to whether you think we get more relevant numbers from [0166717](https://github.com/Restioson/xtra/pull/66/commits/0166717d70d107995318c8d6367bbe4590c671cc). The pattern I was seeing before was just that each n would just be 10x higher than the last, which is exactly as to be expected, so it was not very useful to have the multiple levels without batching like so.

I'll have to look at the benchmarks from first principles. At a glance, having a 10x increasing loop ourselves makes little sense because criterion is already doing warmup loops and stuff.

What might be more interesting is measuring dispatch of messages of different sizes and multiple different messages.

Restioson:
> What might be more interesting is measuring dispatch of messages of different sizes 

I should imagine that the main difference would be in box allocation time. So, there should be some difference between a zst and non zst, but from there we might just be measuring the allocator.


> and multiple different messages.

Could you give a more specific example?

Restioson:
> This is only a very hand-wavy idea but having a benchmark with just dispatching a single message might lead to too many optimizations

I see what you mean. However, I wonder if it actually will change the xtra code paths or not - ultimately, we want to measure the overhead xtra produces, as opposed to the message handler itself. I *think*, therefore, that the dynamic dispatch would prevent it from inlining too much. Maybe the branch predictor would be able to guess that it's the same kind of message each time, but if compiling to a vtable I'm not sure that would happen? You're right though - maybe we could try alternate two different messages and see if it makes much of a difference. 

Would it be okay to do these in a followup PR however, so that at least we have the criterion infra to support further benchmark exploration?

Restioson:
Okay, awesome! Will merge this now then

PR #67, opened 2022-04-10
Add a debug implementation for Address

### LINKED ISSUE #54: Could `Address` implement `fmt::Debug`?
Sometimes it is useful for messages to implement `fmt::Debug` but if they contain `Address`es, that is not possible without a manual impl. Could `Address` implement `Debug`?

Not sure if we can print something useful but the counter value could be interesting for example!

Restioson:
This does not address MessageChannel as I just haven't looked into it yet. I'll look into it before this gets merged, though. Closes #54.

- [x] Implement Debug on RefCounter and show reference counts
- [x] Implement for MessageChannel
- [x] Add test

### DIFF
--- a/examples/basic_wasm_bindgen/src/lib.rs
+++ b/examples/basic_wasm_bindgen/src/lib.rs
@@ -5,15 +5,19 @@ use xtra::spawn::WasmBindgen;
 
 struct Echoer;
 
-impl Actor for Echoer {}
+#[async_trait]
+impl Actor for Echoer {
+    type Stop = ();
+
+    async fn stopped(self) {}
+}
 
 struct Echo(String);
-impl Message for Echo {
-    type Result = String;
-}
 
 #[async_trait]
 impl Handler<Echo> for Echoer {
+    type Return = String;
+
     async fn handle(&mut self, echo: Echo, _ctx: &mut Context<Self>) -> String {
         echo.0
     }
--- a/examples/basic_wasm_bindgen/src/lib.rs
+++ b/examples/basic_wasm_bindgen/src/lib.rs
@@ -9,7 +9,7 @@ struct Echoer;
 impl Actor for Echoer {
     type Stop = ();
 
-    async fn stopped(self) {}
+    async fn stopped(self) -> Self::Stop {}
 }
 
 struct Echo(String);
--- a/src/address.rs
+++ b/src/address.rs
@@ -1,7 +1,7 @@
 //! An address to an actor is a way to send it a message. An address allows an actor to be sent any
 //! kind of message that it can receive.
 
-use std::fmt::{self, Display, Formatter};
+use std::fmt::{self, Debug, Display, Formatter};
 use std::future::Future;
 use std::mem;
 use std::pin::Pin;
@@ -119,6 +119,12 @@ impl Error for Disconnected {}
 pub struct Address<A, Rc: RefCounter = Strong> {
     pub(crate) sender: Sender<AddressMessage<A>>,
     pub(crate) ref_counter: Rc,
+}
+
+impl<A, Rc: RefCounter> Debug for Address<A, Rc> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "Address<{}, {}>", std::any::type_name::<A>(), std::any::type_name::<Rc>())
+    }
 }
 
 /// A `WeakAddress` is a reference to an actor through which [`Message`s](../trait.Message.html) can be
--- a/src/address.rs
+++ b/src/address.rs
@@ -46,10 +46,7 @@ pub struct Address<A, Rc: RefCounter = Strong> {
 
 impl<A, Rc: RefCounter> Debug for Address<A, Rc> {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        let a = std::any::type_name::<A>();
-        let rc = std::any::type_name::<Rc>();
-
-        f.debug_struct(&format!("Address<{}, {}>", a, rc))
+        f.debug_struct(&format!("Address<{}>", std::any::type_name::<A>()))
             .field("ref_counter", &self.ref_counter)
             .finish()
     }
--- a/src/address.rs
+++ b/src/address.rs
@@ -184,7 +184,7 @@ impl<A, Rc: RefCounter> Address<A, Rc> {
     /// })
     /// ```
     pub fn is_connected(&self) -> bool {
-        self.ref_counter.is_connected()
+        !self.sender.is_disconnected()
     }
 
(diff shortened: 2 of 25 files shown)


thomaseizinger (commented):
I think that is a good starting point. Implementing Debug on RefCounter could also be interesting to see the actual refcount.

Restioson:
That makes sense

Restioson:
Is this what you were thinking of when it comes to the reference count debugging @thomaseizinger ?

thomaseizinger:
Yeah this looks good! Perhaps we can add a test that asserts the Debug implementation? This would make it even easier to review and we make sure it is not broken :)

src/address.rs
```diff
@@ -44,6 +44,17 @@ pub struct Address<A, Rc: RefCounter = Strong> {
     pub(crate) ref_counter: Rc,
 }
 
+impl<A, Rc: RefCounter> Debug for Address<A, Rc> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        let a = std::any::type_name::<A>();
+        let rc = std::any::type_name::<Rc>();
+
+        f.debug_struct(&format!("Address<{}, {}>", a, rc))
+            .field("ref_counter", &self.ref_counter)
```
thomaseizinger, line 53: The strong/weak part is duplicated here, isn't it? I think we can remove it from the `Address` type itself and only have it as part of the `ref_counter` field?

Restioson: That is technically true. I mostly kept it to show the full type since  Address<A> might seem to imply strong as that's the default

Restioson:
Sounds like a good plan

src/address.rs
thomaseizinger: Fair enough!
Let's see what it looks like when printed but it would be nice to reduce the duplication one way or another :)

Restioson:
I've added a test for this now. One remaining unresolved question is whether to report the weak count strictly accurately (i.e number of weak references, incl. the one kept inside Context), or to make it count the number of weak addresses (statically one less than the strict count).

thomaseizinger:
> I've added a test for this now. One remaining unresolved question is whether to report the weak count strictly accurately (i.e number of weak references, incl. the one kept inside Context), or to make it count the number of weak addresses (statically one less than the strict count).

I think reporting the actual number of addresses is good enough. I see very little foot-gun potential here :)

tests/basic.rs
```diff
@@ -273,3 +273,25 @@ async fn address_send_exercises_backpressure() {
         .now_or_never()
         .expect("be able to queue another message because the mailbox is empty again");
 }
+
+#[tokio::test]
+async fn address_debug() {
```
thomaseizinger, line 278: Technically, this doesn't need to be an `async` test.

tests/basic.rs
```diff
@@ -273,3 +273,25 @@ async fn address_send_exercises_backpressure() {
...
+    let (addr1, _ctx): (Address<_>, Context<Greeter>) = Context::new(None);
+
+    let addr2 = addr1.clone();
+    let weak_addr = addr2.downgrade();
+
+    assert_eq!(
+        format!("{:?}", addr1),
+        "Address<basic::Greeter, xtra::refcount::Strong> { \
+        ref_counter: Strong { connected: true, strong_count: 2, weak_count: 2 } }"
```
thomaseizinger, line 287: If you use raw strings (`r#""#`), you can omit the escaping of the newlines.

thomaseizinger: I just realized that this isn't even a multiline string. Is this just on two lines because of the formatter?

tests/basic.rs
```diff
@@ -273,3 +273,25 @@ async fn address_send_exercises_backpressure() {
...
+    );
+
+    assert_eq!(format!("{:?}", addr1), format!("{:?}", addr2));
+
+    assert_eq!(
+        format!("{:?}", weak_addr),
+        "Address<basic::Greeter, xtra::refcount::Weak> { \
+        ref_counter: Weak { connected: true, strong_count: 2, weak_count: 2 } }"
```
thomaseizinger, line 295: Seeing this, I'd suggest to simplify the representation to:

```rust
"Address<basic::Greeter> { ref_counter: Weak { connected: true, strong_count: 2, weak_count: 2 } }"
```
or

```rust
"Address<basic::Greeter, xtra::refcount::Weak> { connected: true, strong_count: 2, weak_count: 2 }"
```

tests/basic.rs
Restioson: Good point

tests/basic.rs
Restioson: It's slightly more readable broken across two lines, otherwise it tends to go off the monitor

tests/basic.rs
Restioson: This makes sense. I think I prefer the first probably since it'll show which variant of ref counter is used if it's erased through Either

tests/basic.rs
```diff
@@ -273,3 +273,25 @@ async fn address_send_exercises_backpressure() {
         .now_or_never()
         .expect("be able to queue another message because the mailbox is empty again");
 }
+
+#[test]
+fn address_debug() {
+    let (addr1, _ctx): (Address<_>, Context<Greeter>) = Context::new(None);
```
thomaseizinger, line 279: You could avoid the long type hint with turbo-fish syntax but that is a style preference :)


```suggestion
    let (addr1, _ctx) = Context::<Greeter>::new(None);
```

thomaseizinger (approved):
Implementation looks good to me :)

examples/basic_wasm_bindgen/src/lib.rs
```diff
@@ -5,15 +5,19 @@ use xtra::spawn::WasmBindgen;
...
-    type Result = String;
+    async fn stopped(self) {}
 }
 
+struct Echo(String);
+
 #[async_trait]
 impl Handler<Echo> for Echoer {
+    type Return = String;
+
```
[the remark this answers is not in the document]
Restioson: Yea... I had thought that I'd already rebased it on master but apparently not

src/refcount.rs
```diff
@@ -35,6 +44,11 @@ impl Strong {
         }
     }
 
+    fn weak_count(&self) -> usize {
+        let _lock = self.lock.read().unwrap();
```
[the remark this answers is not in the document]
Restioson: That's true - especially since it's only used in a debug implementation and doesn't require upgrading, downgrading, or dropping anything

Restioson:
Going to go ahead and merge this now if CI passes

Restioson:
I'm pretty confident that CI will pass once #80 lands, since the current failure has nothing to do with any code changes in this branch, so I am going to go ahead and merge this now.

PR #69, opened 2022-04-15
Improve CI coverage and fix doc tests

thomaseizinger:
It happened multiple times recently that CI did not catch all compile errors in docs or other tests. This patch-set (hopefully) fixes those.


Restioson:
Thanks for this! I'm not really sure how github actions works myself (it was somebody else who set it up), so I'll take your word (and the fact that tests pass) that this is the right way to go about this

PR #70, opened 2022-04-16
Remove `Actor::stopping` and `KeepRunning`

thomaseizinger:
This builds on top of https://github.com/Restioson/xtra/pull/73 and should only be reviewed from https://github.com/Restioson/xtra/pull/70/commits/2f82510e2b82a1e3c4b51053beb866ad3521a0f1 onwards.

Marked as draft because of that.


thomaseizinger:
This might be obsolete with the discovery that I made while writing https://github.com/Restioson/xtra/issues/71. `Address::attach_stream` can actually be removed in favor of [`Stream::forward`](https://docs.rs/futures/0.3.21/futures/stream/trait.StreamExt.html#method.forward) and converting the `Address`/`MessageChannel` into a `Sink`.

Should we encourage users to do that instead? We could document that by adding an example.

Restioson:
I'm okay with the concept of this change. End-users can replace
```rust
fn stopping(...) -> KeepRunning {
    if thing {
        KeepRunning::KeepRunning
    } else {
       KeepRunning::StopAll
    }
}

// somewhere...
ctx.stop();
```

with

```rust
// somewhere...
if !thing {
   ctx.stop();
}
```

PR #72, opened 2022-04-16
Fix code example in readme

thomaseizinger:


Restioson:
Thanks, forgot about this as it isn't included in doc tests!

Issue #64, opened 2022-04-09
Remove `stopping` and the entire `KeepRunning` construct


thomaseizinger:
Whilst working on https://github.com/Restioson/xtra/pull/51, I came to the conclusion that I'd like to propose to remove the `stopping` function and with it the concept of `KeepRunning`.

The reasons why I consider it problematic are:

- With `KeepRunning::StopAll`, we may terminate actors without them being able to prevent that. It seems weird that a single actor can intervene shutdown (when initiated via `Context#stop`) but other actors cannot.
- IMO, an actor should not be able to halt a shutdown when initiated via `Context#stop`. It seems much cleaner to use a supervisor pattern (see https://github.com/itchysats/itchysats/blob/master/xtras/src/supervisor.rs for example) to _restart_ an actor when it was shut down. With the recent addition of `Actor#Stop`, this is actually super easy and clean to implement.

The current features of `stopping` and `KeepRunning` can still be achieved through other means:

- Shutting down all actors in a `Context` can be done with a custom message and the `notify_all` function
- Preventing an actor from shutdown is not possible but it can be restarted with a supervisor (see above)

Removing `stopping` and `KeepRunning` would IMO also simplify a lot of the state management in `Context` and make https://github.com/Restioson/xtra/pull/51 easier to implement / finish. What do you think?

Restioson:
This does make sense to me, I think. I really like how this simplifies control flow and general usage - less ways to do the same thing. You can emulate preventing an actor from shutdown by adding a `ShutdownMaybe` message which then checks if the actor should be shut down, and calls `Context::stop` if required.

Restioson:
This would supercede and close #58, right?

thomaseizinger:
> This would supercede and close #58, right?

Yes.

thomaseizinger:
> This does make sense to me, I think. I really like how this simplifies control flow and general usage - less ways to do the same thing. You can emulate preventing an actor from shutdown from `Actor::stopping` by adding a `ShutdownMaybe` message which then checks if the actor should be shut down, and calls `Context::stop` if required. 

Yep, I had a similar thought. I think moving more of this functionality into messages is a good thing.

I've also been contemplating the idea of making the `started` lifecycle hook a message but I am not fully sure about that yet :)

Restioson:
How would replacing started work? Would each address of the actor be made to send a start message?

thomaseizinger:
[...]

Issue #71, opened 2022-04-16
Discussion: Smaller and safer API surface


thomaseizinger:
`xtra` wants to be a _tiny, fast, and safe actor framework_.

After working with it for a while, I think this is mostly true but I also think we can do better, mostly on the _tiny_ and _safe_ part. I'd assume there are also performance gains to be made but I know where little about that so I am going to leave that for people with more experience.

At the moment, `xtra`'s API is already fairly small which I'd consider to be a good thing. However, there are some parts that I'd consider to be inconsistent and where I think we could do better in terms of composability. Additionally, there are some feature where I am not sure whether they need to necessarily be in the core `xtra` crate.

1. `do_send`, `do_send_async`, `send`.

We have 3 APIs for sending messages on an `Address`. In our project, we [ban](https://github.com/itchysats/itchysats/blob/6ef58b2b5fbce8535c41a394e748fb45e1a9d8dd/clippy.toml#L2-L4) the `do_send` ones. We consider them to be foot-guns because they silently drop return values which could be a `Result` and thus, things might be failing without one noticing. I'd like to propose to remove those from the public API. Asynchronous sending can be achieved by spawning the `SendFuture` into an executor. The fact that our underlying channels have a dedicated _async` way of sending messages should be hidden IMO.

2. `Address::attach_stream`

A nice convenience method that we used a fair bit initially but quickly realized, that it is not as useful as it initially seems, primarily because it does not give you a way of dealing with the return value of the `Handler`. I think this function can be removed or moved into some kind of `xtra-ext` crate. It builds on top of the `Address::send` API and could thus easily live in another crate.

3. `Context::notify_interval`

We have found this function to be problematic in two ways:

a. It sends messages asynchronously to the actor, thus potentially piling up a number of messages if the actor cannot process them quickly enough.
b. It does not allow for an error handling policy. We have many handlers that are invoked on an interval that can fail (background sync jobs) but their failure should not abort the loop but instead log a message.

Given that the function really only composes `send`, I think we should try and move it out of the core `xtra` crate and provide ways for users to easily create their own "send on an interval" loop. Perhaps this can be combined with `Address::attach_stream`? An stream of intervals that maps to messages would achieve the same outcome I think.

4. `Context::notify_after`

Similar to `notify_interval` this function only composes `send` and thus, I think we could lift it into a separate crate, minimizing the API surface of `Context`.

5. `Context::{notify,notify_all}`

[...]

PR #73, opened 2022-04-17
Make `Address` implement `Sink` for `Handler`s that return `()`

thomaseizinger:
This is an alternative to https://github.com/Restioson/xtra/pull/70 and would therefore also help with https://github.com/Restioson/xtra/issues/64.

I read through https://github.com/Restioson/xtra/pull/48 which originally added a test that I had to remove because the API no longer exists now.

`StreamExt::forward` has (unfortunately) a different behaviour to what was "fixed" in https://github.com/Restioson/xtra/pull/48. In particular, it behaves the same as the original implementation and the current `Sink` implementation: It does **not** immediately return if the actor gets stopped but would only fail on the next item in the stream.

I think this is actually the correct behaviour. One could argue that the system is actually not broken if a stream that does not produce any items tries to forward to an actor that is stopped.

Let me know what you think.


thomaseizinger:
@Restioson Based on your [comment](https://github.com/Restioson/xtra/issues/71#issuecomment-1106492397), would you accept merging this if I drop the commits that remove `Address::attach_stream` and `Context::notify_interval`?

Without those commits, this PR is only a refactoring that removes the separate `AddressSink` but retains all functionality otherwise and should thus be non-controversial to land given that it deletes a fair bit of code?

Restioson:
Sorry for letting this slip my mind too - I think overall the main issue is the tradeoff between code size/maintainability, and the size of the Address struct as a whole. Adding a sink as a field could increase it from the current 16 bytes (IIRC) to maybe 24 or 32 with padding, which could have implications if lists of actor addresses are kept. However, I'm not sure how much of an issue this will be considering that actors can share addresses and you don't need to implement message stealing yourself. What is your opinion on this?

thomaseizinger:
> Adding a sink as a field could increase it from the current 16 bytes (IIRC) to maybe 24 or 32 with padding, which could have implications if lists of actor addresses are kept.

Yes, I thought about this but I figured that the added convenience and simplified codebase is worth it.

For this to seriously manifest, we would have to have _a lot_ of addresses in a list no?

I'd err on the side of perhaps optimising the other way in case the size of the struct ever becomes an issue: Adding a `SmallAddress` or something that does not implement `Sink`.

Additionally, I value the API over the implementation details. I am sure if we'd think hard about it, we could optimise the size of `Address` :)

thomaseizinger:
Rebased to resolve conflicts, made no other changes.

Restioson:
I've opened https://github.com/zesterer/flume/pull/102 in this interest - I will prototype something on the xtra side based on my fork shortly.

thomaseizinger:
#81 is much better than this! Happy to close this in favor :)

Restioson:
> Additionally, I value the API over the implementation details. I am sure if we'd think hard about it, we could optimise the size of Address :)

Speaking of this, if #85 works out, then it should be possible to have Address be exactly pointer sized, with atomic counters for strong and weak to replace the dedicated ref counter atomic (I will elaborate on this in #85 later). However, this is only if this change is reverted - Address would no longer be a Sink. However, I think it might be possible to expose a `fn address(&self) -> &Address` which would remove the `MessageSink` trait and with that a lot of maintenance burden. Of course, #85 itself is also far more complicated but I believe it makes sense for the library and I will elaborate on those reasons there, and we can discuss it if you are interested in doing so :)

Restioson:
Would there be any issue with adding a method to convey the method to a sink?

thomaseizinger:
> This can theoretically be done in an extension crate too, if xtra decides to take that route to improve maintainability. Nameability of types aside, it's semantically and functionally equivalent to https://docs.rs/futures/0.3.9/futures/sink/fn.unfold.html using Address::send

Fair point!

I think it would be good to try and distill the API down to the absolute minimum so happy to drop the `Sink` impl for now if that is what is necessary for that :)

PR #74, opened 2022-04-17
Unify the `send` API on `Address` and `MessageChannel`

thomaseizinger:
This PR presents a unified API for sending messages to actors. We now only have a single `send` function on both `Address` and `MessageChannel` and we return the same `SendFuture` type in both cases.

The `SendFuture` type has a single, publicly accessible function called `recv_async` which allows users to flip the behaviour of `SendFuture` from synchronously awaiting the handler's result to returning a `Receiver` that can be awaited separately.

This allows users to _queue_ messages into an actor's mailbox without awaiting the result. As part of merging the two existing `SendFuture` types, we now also provide a consistent "backpressure" story: A `SendFuture` that is flipped to receiving the result asynchronously will only complete once the message has been successfully queued into the actor's mailbox. I've added an example that demonstrates this and also a test that make sure we don't break this functionality in the future.

### DIFF
--- a/benches/throughput.rs
+++ b/benches/throughput.rs
@@ -47,7 +47,7 @@ fn throughput(c: &mut Criterion) {
             |b, &num_messages| {
                 b.to_async(&runtime).iter(|| async {
                     for _ in 0..num_messages - 1 {
-                        address.send(IncrementZst).split_receiver().await.unwrap();
+                        let _ = address.send(IncrementZst).split_receiver().await;
                     }
 
                     address.send(IncrementZst).await.unwrap()
--- a/benches/throughput.rs
+++ b/benches/throughput.rs
@@ -47,7 +47,7 @@ fn throughput(c: &mut Criterion) {
             |b, &num_messages| {
                 b.to_async(&runtime).iter(|| async {
                     for _ in 0..num_messages - 1 {
-                        address.send(IncrementZst).await.unwrap();
+                        address.send(IncrementZst).recv_async().await.unwrap();
                     }
 
                     address.send(IncrementZst).await.unwrap()
--- a/benches/throughput.rs
+++ b/benches/throughput.rs
@@ -47,7 +47,7 @@ fn throughput(c: &mut Criterion) {
             |b, &num_messages| {
                 b.to_async(&runtime).iter(|| async {
                     for _ in 0..num_messages - 1 {
-                        address.do_send(IncrementZst).unwrap();
+                        address.send(IncrementZst).await.unwrap();
                     }
 
                     address.send(IncrementZst).await.unwrap()
--- a/benches/throughput.rs
+++ b/benches/throughput.rs
@@ -47,7 +47,7 @@ fn throughput(c: &mut Criterion) {
             |b, &num_messages| {
                 b.to_async(&runtime).iter(|| async {
                     for _ in 0..num_messages - 1 {
-                        address.send(IncrementZst).recv_async().await.unwrap();
+                        address.send(IncrementZst).split_receiver().await.unwrap();
                     }
 
                     address.send(IncrementZst).await.unwrap()
--- a/src/address.rs
+++ b/src/address.rs
@@ -133,10 +133,12 @@ impl<A, Rc: RefCounter> Address<A, Rc> {
         }
     }
 
-    // TODO: Revise these docs.
-    /// Send a [`Message`](../trait.Message.html) to the actor and asynchronously wait for a response. If this
-    /// returns `Err(Disconnected)`, then the actor is stopped and not accepting messages. Like most
-    /// futures, this must be polled to actually send the message.
+    /// Send a message to the actor.
+    ///
+    /// The actor must implement [`Handler<Message>`] for this to work.
+    ///
+    /// This function returns a [`Future`](SendFuture) that resolves to the [`Return`](crate::Handler::Return) value of the handler.
+    /// The [`SendFuture`] will resolve to [`Err(Disconnected)`] in case the actor is stopped and not accepting messages.
     #[allow(clippy::type_complexity)] // TODO: Actually fix this
     pub fn send<M>(
         &self,
--- a/src/address.rs
+++ b/src/address.rs
@@ -79,7 +79,9 @@ impl<A: Actor, R> Future for SendFuture<A, R, ReceiveSync> {
     fn poll(self: Pin<&mut Self>, ctx: &mut Context) -> Poll<Self::Output> {
         let this = self.get_mut();
         let (poll, new) = match mem::take(&mut this.inner) {
-            old @ SendFutureInner::Disconnected => (Poll::Ready(Err(Disconnected)), old),
+            SendFutureInner::Disconnected => {
+                (Poll::Ready(Err(Disconnected)), SendFutureInner::Done)
+            }
             SendFutureInner::Sending(mut tx, mut rx) => {
                 if tx.poll_unpin(ctx).is_ready() {
                     (poll_rx(&mut rx, ctx), SendFutureInner::Receiving(rx))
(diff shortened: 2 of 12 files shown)


zesterer:
Could you give a few examples of patterns enabled by this change?

thomaseizinger:
Well, it is mostly a refactoring really. Did you have a chance to look at the two tests I added that demonstrate the usage?

But I'll take your feedback and add some examples and documentation :)

thomaseizinger:
@Restioson @zesterer This is ready for a full-review.

The PR description is up-to-date. I wouldn't bother with a patch-by-patch review but just look at the entire diff at once :)
Looking forward to your feedback!

zesterer:
I'm really liking the look of these changes! I'm not sure how well I'd be able to comment on the implementation without more knowledge of the crate's internals, but this API is definitely much closer to the intuitive behaviour I'd expect from an actor library. Thanks for you work on this!

Restioson:
Awesome - I will try take a look next week as I am away for this weekend

src/send_future.rs
```diff
@@ -0,0 +1,180 @@
...
+
+impl<R, F> SendFuture<R, F, ReceiveSync> {
+    pub(crate) fn disconnected() -> Self {
+        Self {
+            inner: SendFutureInner::Disconnected,
+            phantom: PhantomData,
+        }
+    }
+
+    /// Toggle this future to only queue the message into the actor's mailbox.
```
thomaseizinger, line 46:
```suggestion
    /// Toggle this future to receive the return value asynchronously.
```

src/send_future.rs
```diff
@@ -0,0 +1,180 @@
...
+    ///
+    /// Calling this function will change the [`Output`](Future::Output) of this [`Future`] from [`Handler::Return`](crate::Handler::Return) to [`Receiver<Handler::Return>`](Receiver<crate::Handler::Return>).
+    ///
+    /// A [`Receiver`] is a [`Future`] itself that can be used to await the execution of the [`Handler`](crate::Handler) separately from the queuing of the message.
+    pub fn recv_async(self) -> SendFuture<R, F, ReceiveAsync> {
```
thomaseizinger, line 51: TODO: This is probably a good place for a doc test that explains what can be done with this.

tests/basic.rs
```diff
@@ -185,3 +189,86 @@ impl Handler<Stop> for ActorReturningStopSelf {
...
+        tokio::time::sleep(duration).await
+    }
+}
+
+#[tokio::test]
+async fn receiving_async_on_address_returns_immediately_after_dispatch() {
+    let address = LongRunningHandler.create(None).spawn_global();
+
+    let send_future = address.send(Duration::from_secs(3)).recv_async();
+    let handler_future = send_future.now_or_never().unwrap(); // Dispatch should be immediate on first poll!
```
thomaseizinger, line 216:
```suggestion
    let handler_future = send_future.now_or_never().expect("Dispatch should be immediate on first poll");
```

tests/basic.rs
```diff
@@ -185,3 +189,86 @@ impl Handler<Stop> for ActorReturningStopSelf {
...
+    handler_future.await.unwrap();
+}
+
+#[tokio::test]
+async fn receiving_async_on_message_channel_returns_immediately_after_dispatch() {
+    let address = LongRunningHandler.create(None).spawn_global();
+    let channel = StrongMessageChannel::clone_channel(&address);
+
+    let send_future = channel.send(Duration::from_secs(3)).recv_async();
+    let handler_future = send_future.now_or_never().unwrap(); // Dispatch should be immediate on first poll!
```
thomaseizinger, line 227:
```suggestion
    let handler_future = send_future.now_or_never().expect(Dispatch should be immediate on first poll");
```

src/send_future.rs
```diff
@@ -0,0 +1,180 @@
...
+            phantom: PhantomData,
+        }
+    }
+
+    /// Toggle this future to receive the return value asynchronously.
+    ///
+    /// Calling this function will change the [`Output`](Future::Output) of this [`Future`] from [`Handler::Return`](crate::Handler::Return) to [`Receiver<Handler::Return>`](Receiver<crate::Handler::Return>).
+    ///
+    /// A [`Receiver`] is a [`Future`] itself that can be used to await the execution of the [`Handler`](crate::Handler) separately from the queuing of the message.
+    pub fn recv_async(self) -> SendFuture<R, F, ReceiveAsync> {
```
Restioson, line 51: I like the concept of moving out the change in receiving behavior from `do_` to the `send` method itself with this kind of builder-style method. I wonder if there is a better name for it though - this method doesn't make the receiving async (like the name could imply), it rather just makes it happen separately from the send action. Ultimately, I'd just be worried the name would imply that `send` blocks to receive, which it does not

src/address.rs
```diff
@@ -207,53 +133,21 @@ impl<A, Rc: RefCounter> Address<A, Rc> {
...
-
-    /// Send a [`Message`](../trait.Message.html) to the actor and asynchronously wait for a response. If this
-    /// returns `Err(Disconnected)`, then the actor is stopped and not accepting messages. Like most
-    /// futures, this must be polled to actually send the message.
-    pub fn send<M>(&self, message: M) -> SendFuture<A, <A as Handler<M>>::Return>
+    /// Send a message to the actor.
+    ///
+    /// The actor must implement [`Handler<Message>`] for this to work.
+    ///
+    /// This function returns a [`Future`](SendFuture) that resolves to the [`Return`](crate::Handler::Return) value of the handler.
```
Restioson, line 140: Just a side note - I hadn't realised how powerful the new rustdoc linking was 😄 Most of the links in xtra predate its stabilisation

src/send_future.rs
```diff
@@ -0,0 +1,180 @@
...
+        SendFuture {
+            inner: self.inner,
+            phantom: PhantomData,
+        }
+    }
+}
+
+impl<R> SendFuture<R, BoxFuture<'static, Receiver<R>>, ReceiveSync> {
+    pub(crate) fn sending_boxed<F>(send_fut: F) -> Self
```
Restioson, line 60: Just to check - this is used for message channels, whilst sending_named is to specific actor types?

benches/throughput.rs
```diff
@@ -47,7 +47,7 @@ fn throughput(c: &mut Criterion) {
             |b, &num_messages| {
                 b.to_async(&runtime).iter(|| async {
                     for _ in 0..num_messages - 1 {
-                        address.do_send(IncrementZst).unwrap();
+                        address.send(IncrementZst).await.unwrap();
```
Restioson, line 50: I think this should probably be
```rust
let _ = address.send(IncrementZst).recv_async().await;
```
Otherwise, the benchmark would be the same no matter the batch size of `num_messages`, would it not?

src/context.rs
```diff
@@ -422,7 +422,7 @@ impl<A: Actor> Context<A> {
                 let delay = Delay::new(duration);
                 match future::select(delay, &mut stopped).await {
                     Either::Left(_) => {
-                        if addr.do_send(constructor()).is_err() {
+                        if addr.send(constructor()).await.is_err() {
```
Restioson, line 425: This (as well as the change on L458) would change how the methods work w.r.t backpressure. Whilst the old behaviour was not explicitly documented, it might be worth noting in the doc comment now that it will be changing from how it used to work.

Restioson: This, however, would be irrelevant if split out into an extension crate

src/send_future.rs
```diff
@@ -0,0 +1,180 @@
...
+///
+/// When toggled to receive the return value asynchronously, this future will resolve once the message is successfully queued into the actor's mailbox. If the actors mailbox is bounded, this future will yield `Pending` until there is space in the mailbox. This allows an actor to exercise backpressure on its users.
+#[must_use = "Futures do nothing unless polled"]
+pub struct SendFuture<R, F, TRecvSyncMarker> {
+    inner: SendFutureInner<R, F>,
+    phantom: PhantomData<TRecvSyncMarker>,
+}
+
+/// Marker type for receiving the return value of a [`Handler`](crate::Handler) **synchronously**.
+pub enum ReceiveSync {}
```
Restioson, line 26: When you say synchronously, I assume you mean as part of the same future as the `send()`, so still not with blocking channel `recv`?

src/send_future.rs
thomaseizinger: Correct!

src/send_future.rs
thomaseizinger: Yes, synchronously here means within the same future/task!

thomaseizinger: I think we should document `xtra`-wide that we attempt to never block threads whatsoever so "blocking" always just means "blocking the task".

src/address.rs
thomaseizinger: They are pretty nice yeah :)

src/send_future.rs
thomaseizinger: > would imply that send blocks to receive, which it does not

Well, it does block the _task_ until the return value is ready. The mix of terminology here can be confusing though, I agree!

thomaseizinger: > this method doesn't make the receiving async (like the name could imply), it rather just makes it happen separately from the send action.

When ignoring the specific semantics of `async/await` in Rust, the usage of "asynchronously" here is still correct I believe. It is like having:

1. An HTTP request which receives the response synchronously.
2. A "fire-and-forget" HTTP request that returns you a link from where you can fetch the result eventually.

I am happy to rename this to something else because the naming is clearly confusing due to the overloading of the term `asynchronously` with Rust's specific semantics.

A few other names I can think of are:

- `split_receiver`
- `into_receiver`
- `recv_separately`
- `queue_only` (because the future will resolve once the message is queued into the mailbox)

@zesterer Any opinion on the naming of this API?

benches/throughput.rs
thomaseizinger: I didn't pay too much attention to this to be honest but yes, this makes total sense!

src/send_future.rs
thomaseizinger: I had a go at renaming the function to `split_receiver` in https://github.com/Restioson/xtra/pull/74/commits/49e2f56de9a41dc2e45809e800898d97b0247467. Let me know what you think!

Restioson: That sounds nice to me I think - no more clash with Rust's sync vs async terminology :)

Restioson:
Is there anything else to be done for this or is it ready for merge?

thomaseizinger:
Are all bigger concerns like naming, design etc dealt with from your perspective?

If yes, then I'll do another pass and tidy up smaller things, make sure all the doc comments are good etc.

Restioson:
I think so, yes. Overall it was mostly naming things - I like the general structure of the changes a lot.

thomaseizinger:
> Hi, just checking if everything is finalised from your side or is there anything else you'd still like to change?

Thanks the check-in!

I've been a bit busy lately but I'll try to do the clean-up over the next couple of days :)

Restioson:
No problem at all. I've also been busy which is why this has slipped my mind, but since it's consolidation week now, we aren't being flooded with tests and assignments anymore, so I am going to try and make sure that I get up to date on reviews and finish off my PR on debug implementations 😄

thomaseizinger:
Okay, this is ready! :)

PR #76, opened 2022-04-28
Type-safe actor lifecycle

thomaseizinger:
Currently, we have a funny condition within `run` where actors can theoretically call `Context::stop` from within the `started` function. A code-comment currently hints that this idea is a bit silly.

By using a separate type-parameter, we can indicate the "life-cycle stage" of the actor. Through this, we can make it impossible to call `Context::stop` as part of the actor's `started` lifecycle callback.

To still allow for a fallible `started` function, we change the type-signature to return `Result<(), Self::Stop>`.

If this idea is welcome, I can polish the code a bit more. Names for life-cycle stages are subject to bike-shedding. At the moment, it is also not possible to call `handle_while` and other `Context` functions as part of `started`. This decision may need to be revisited.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,6 +1,7 @@
 use std::fmt;
 use std::fmt::{Display, Formatter};
 use std::future::Future;
+use std::marker::PhantomData;
 use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 
@@ -19,7 +20,7 @@ use crate::{Actor, Address, Handler, KeepRunning};
 
 /// `Context` is used to control how the actor is managed and to get the actor's address from inside
 /// of a message handler.
-pub struct Context<A> {
+pub struct Context<A, TStage = Running> {
     /// Whether the actor is running. It is changed by the `stop` method as a flag to the `ActorManager`
     /// for it to call the `stopping` method on the actor
     running: RunningState,
@@ -39,7 +40,15 @@ pub struct Context<A> {
     /// to shutdown the tasks as soon as the context stops.
     #[cfg_attr(not(feature = "timing"), allow(dead_code))]
     drop_notifier: DropNotifier,
+
+    stage: PhantomData<TStage>,
 }
+
+/// Represents the `Starting` stage of an actor.
+pub enum Starting {}
+
+/// Represents the `Running` stage of an actor.
+pub enum Running {}
 
 #[derive(Eq, PartialEq, Copy, Clone)]
 enum RunningState {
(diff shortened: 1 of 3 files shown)


src/context.rs
```diff
@@ -19,7 +20,7 @@ use crate::{Actor, Address, Handler, KeepRunning};
 
 /// `Context` is used to control how the actor is managed and to get the actor's address from inside
 /// of a message handler.
-pub struct Context<A> {
+pub struct Context<A, TStage = Running> {
```
thomaseizinger, line 23: Currently setting this as the default to reduce the diff. We may want to be explicit about it in the final version.

Restioson: Sounds sensible - what does the T in TStage stand for exactly, though?

src/context.rs
```diff
@@ -123,24 +143,90 @@ impl<A: Actor> Context<A> {
             broadcast_receiver,
```
Restioson, line 143: Should this function (attach) be only in the starting phase? It might make sense to add extra actors dynamically if the workload increases, for instance

Restioson (commented):
Sorry for not getting to this until now - it totally slipped my mind. Broadly, I think this change make sense, but there's just one or two things I think should be clarified before it can be merged. Additionally, do you think this should be merged before or after the changes to send, or does it not really matter?

src/context.rs
thomaseizinger: It stands for "Type" and is a convention I sometimes use for type parameters that have more descriptive names.

I don't feel strongly about it though, happy to rename it to whatever :)

src/context.rs
thomaseizinger: I am unsure about that as I've never used multiple actors on a single `Context`.

Technically, I think we can expose the function for any lifecycle stage.

Restioson (approved):
This looks good to me - are you happy with it?

thomaseizinger:
> This looks good to me, just aside from having `attach` in `Running` too - are you happy with it? I'd be happy to merge as is and make the last change on master.

I'd like to remove the type parameter default first, plus I have an idea on how to also use the type parameter for making the stopping phase more type-safe. That will be more important for custom event loops though so if you are keen on getting this in, feel free to go ahead :)

Having `attach` in `Running` too makes sense to me!

Restioson:
> I'd like to remove the type parameter default first

Do you mean defaulting `TStage` to `Running`?

Restioson:
Why not, out of interest? I think that most of the time `Context` would be in a `Running` state, due to it being written more in `Handler`s than in `Actor` implementations

thomaseizinger:
Yes fair point. I am mostly auto-generating `Handler` implementations so I wouldn't notice the default much.

Making it obvious to the user that type-system safety is at work here might be nice to also communicate why they can't call `stop_self` in `started` for example.

Restioson:
That's true, but for `started` specifically it will already say `Context<..., Starting>`. Meanwhile, other than `run` (which you already cannot invoke with `&mut Context`), there are no methods missing from `Context<..., Running>`, so no nasty surprises.

thomaseizinger:
> One other thing - could Context\u003C..., Starting> get an address() method that is infallible, as the actor cannot be already stopped in that stage?

I would love that but I don't think it is entirely true. If you instantly drop the address you are getting back from `run`, there won't be any strong ones left.

Restioson:
Ah, unfortunately you are right :(

PR #81, opened 2022-06-09
Addresses are sinks take 2

Restioson:
This is essentially #73, but with less data duplication thanks to an upstream change I've proposed in flume.


Restioson:
I think this is ready to merge once https://github.com/zesterer/flume/pull/102 is merged and released.

PR #82, opened 2022-06-10
Remove stopping

Restioson:
Alternative to #70 off of master. KeepRunning was kept for `attach_stream`, which can be removed or changed in a followup PR. Also, the behaviour of stop_all has changed (see [441ce14](https://github.com/Restioson/xtra/pull/82/commits/441ce1402b1ac43408b1808e4ed3bc50a4c8247c)). A bug with the various implementations of `is_connected` was fixed too. I also changed the behaviour of `notify_interval` - this would previously be stopped when the actor which originally called `notify_interval` was dropped, which seemed incorrect to me given that `notify_interval` actually sends to all actors on the address.


src/context.rs
```diff
@@ -127,22 +121,14 @@ impl<A: Actor> Context<A> {
...
 
-    /// Stop the actor as soon as it has finished processing current message. This will mean that the
-    /// [`Actor::stopping`](trait.Actor.html#method.stopping) method will be called.
-    pub fn stop(&mut self) {
-        self.running = RunningState::Stopping;
+    /// Stop this actor as soon as it has finished processing current message. This means that the
+    /// [`Actor::stopped`] method will be called.
+    pub fn stop_self(&mut self) {
+        self.running = false;
     }
```
thomaseizinger, line 128: This is a breaking change that I am not sure is necessary?

There is no semantic change here, is there? People relying on `stopping` will already notice that things have changed.

thomaseizinger (commented):
Thanks for pulling this out of #73 :)

src/context.rs
Restioson: It's true that those relying on `stopping` will already see the change, but since `stop_all` is now becoming public, I thought it might be better to rename this and clarify that it stops only the current actor as opposed to all on the address.

thomaseizinger: I didn't see `stop_all` becoming public, makes total sense then!

Restioson:
Turns out my local IDE test runner wasn't invoking all the tests so I missed that one :sweat:

Restioson:
There has been a bit of a change of behaviour with `stop_all` in [441ce14](https://github.com/Restioson/xtra/pull/82/commits/441ce1402b1ac43408b1808e4ed3bc50a4c8247c).

tests/basic.rs
```diff
@@ -157,33 +167,63 @@ async fn test_stream_cancel_join() {
...
+    tokio::spawn(ctx.attach(ActorStopSelf)); // Context not dropped here
+    address.send(Stop).await.unwrap();
+    tokio::time::sleep(Duration::from_secs(1)).await;
+
+    assert!(address.is_connected());
+    address.send(Stop).await.unwrap();
+
+    tokio::time::sleep(Duration::from_secs(1)).await;
+    assert!(address.is_connected());
+}
```
thomaseizinger, line 208: Is this really a good behaviour? I'd expect `address.is_connected()` to be equivalent to "can I expect that a message that I am about to send will be handled".

Restioson: Essentially, `is_connected` is equivalent to "can I expect that a message I am about to send **may** be handled". Since a free context still lives, this means that at a later time, another actor might be spawned onto that address (using `Context::run` or `Context::attach` and then handle the message.

Strictly speaking, **will** be handled was never a guarantee at all - the actor could stop itself with `Context::stop` before getting to a given message, even though when it was sent, `is_connected` returns true. 

> I'd expect address.is_connected() to be equivalent to "can I expect that a message that I am about to send will be handled".

I believe that this behaviour is currently used by [xtras](https://github.com/itchysats/itchysats/blob/master/xtras/src/supervisor.rs#L17) to implement a supervisor, though I haven't looked into it in great depth yet. The impression I got, though, was that if the address closes as soon as the last *actor* (as opposed to the last *context*) stops, then you can't restart an actor if it crashes using the supervisor pattern as currently implemented. Maybe there is another way to implement the supervisor pattern, though?

thomaseizinger: > > I'd expect address.is_connected() to be equivalent to "can I expect that a message that I am about to send will be handled".
> 
> I believe that this behaviour is currently used by [xtras](https://github.com/itchysats/itchysats/blob/master/xtras/src/supervisor.rs?rgh-link-date=2022-06-14T20%3A59%3A25Z#L17) to implement a supervisor, though I haven't looked into it in great depth yet. The impression I got, though, was that if the address closes as soon as the last _actor_ (as opposed to the last _context_) stops, then you can't restart an actor if it crashes using the supervisor pattern as currently implemented. Maybe there is another way to implement the supervisor pattern, though?

Yep that is how I implemented it then but the implementation also guarantees that there won't ever be more than one actor running on this context.

> Essentially, `is_connected` is equivalent to "can I expect that a message I am about to send **may** be handled". Since a free context still lives, this means that at a later time, another actor might be spawned onto that address (using `Context::run` or `Context::attach` and then handle the message.

This is a bit of a surprising behaviour. I would expect `send` to return a `Disconnected` if no actor is alive and not have the message queued until a new actor is spawned, just because the `Context` is still there? Also, is this even true if a combination of `attach` and `run` is used? I guess as soon as you use `run` once, the context will be gone with the last actor stopping.

Restioson: > Yep that is how I implemented it then but the implementation also guarantees that there won't ever be more than one actor running on this context.

That is true. On the other hand, it's probably a bit counterintuitive to make it shut down the address if 2/2 actors on it stop, versus not shutting down the address if 1/1 actors on it stop.

> I would expect send to return a Disconnected if no actor is alive and not have the message queued until a new actor is spawned.

I am mostly worried that this would send the wrong signal to users that "this address is dead, we can throw it away", which would not be the case.

> just because the Context is still there? 

Most of the time, when an actor stops, the only contexts are running. I think that the only exception to this would be in the case of the supervisor pattern or another time when you specifically relying on this behaviour. Therefore, it probably isn't so surprising from either the supervisor or the sender side - from the supervisor's perspective, the implementation becomes impossible if this behaviour is removed. From the sender side, since it may be restarted to handle a message later, `Disconnected` early gives the wrong impression.

> I guess as soon as you use `run` once, the context will be gone with the last actor stopping.

That is true. Technically, though, you could `attach`, say, 2 actors, and when they stop `run`, consuming the context. I'm not sure why anyone would ever do this, though.

Restioson: So, do you think we can merge it? If this behaviour really shouldn't exist at all, we can discuss this further in another PR removing it, I think, since the only change here is that now `stop_all` does the same thing as `stop_self` run on all actors. Fixing this inconsistency I personally think is a bugfix, as it was counterintuitive and undocumented.

thomaseizinger: Yes. I think the behaviour also makes sense with the current capabilities exposed by `xtra`, i.e. the `attach` method.

Slight OT: I am wondering if there is a better way to implement multiple actors behind one address. Perhaps once we get the inbox reworkings in, creating multiple event loops listening on the same address can be implemented on top and thus with different semantics!

Restioson: > Perhaps once we get the inbox reworkings in, creating multiple event loops listening on the same address can be implemented on top and thus with different semantics!

Once it's done, I'd be interested in seeing how this could be achieved! My first thought is that it might depend rather heavily on the inbox channel semantics, though

PR #83, opened 2022-06-10
Rework handle_while

Restioson:
This PR renames `handle_while` to `join`, fixing a potential issue that would cause slight inefficiency if the actor is stopped during handling. It also adds `select` as a counterpart, which is like `join` but will return early if the actor is stopped.


Restioson:
There's also a minor internal spelling error fix in here which can be extracted if needed.

src/context.rs
```diff
@@ -341,12 +343,115 @@ impl<A: Actor> Context<A> {
         let mut addr_recv = addr_rx.recv_async();
         let mut broadcast_recv = broadcast_rx.recv_async();
 
-        loop {
+        while self.running == RunningState::Running {
+            let (next_msg, unfinished) = {
```
Restioson, line 347: This is currently a lot of code duplication, but hopefully if #85 makes it in it should be able to be simplified quite a lot

Restioson:
Some tests for `join` and `select` might be good, but those probably don't need to be reviewed I'd imagine, so this is ready for review, I think.

src/context.rs
```diff
@@ -319,14 +319,16 @@ impl<A: Actor> Context<A> {
         self.tick(self.recv_once().await, act).await;
     }
 
-    /// Handle any incoming messages for the actor while running a given future.
+    /// Joins on a future by handling all incoming messages whilst polling it. The future will
+    /// always be polled to completion, even if the actor is stopped. If the actor is stopped,
+    /// handling of messages will cease, and only the future will be polled.
```
thomaseizinger, line 324: In other words, this behaves similar to `futures::join` which waits on both futures to complete, right?

src/context.rs
```diff
@@ -341,12 +343,115 @@ impl<A: Actor> Context<A> {
...
+    /// // Will print "Started!", "Goodbye!", "Decided not to keep running", and then "Finally stopping."
+    /// # #[cfg(feature = "with-smol-1")]
+    /// smol::block_on(async {
+    ///     let addr = MyActor.create(None).spawn(&mut xtra::spawn::Smol::Global);
+    ///     assert!(addr.is_connected());
+    ///     assert_eq!(addr.send(Selecting).await, Ok(true)); // Assert that the select did end early
+    /// })
+    ///
+    /// ```
+    pub async fn select<F, R>(&mut self, actor: &mut A, mut fut: F) -> Result<R, F>
```
thomaseizinger, line 436: Perhaps `Either` would be a better return value here? I don't think `Result` is appropriate because the actor stopping before isn't really a failure case.

Additionally, this would then move closer to the `futures::select` API.

src/context.rs
```diff
@@ -341,12 +343,115 @@ impl<A: Actor> Context<A> {
...
+    ///         let send = addr.send(Stop).split_receiver();
+    ///         send.await;
+    ///
+    ///         // Actor is no longer running, so this will return Err, even though the future will
+    ///         // usually never complete.
+    ///         select.await.is_err()
+    ///     }
+    /// }
+    ///
+    /// // Will print "Started!", "Goodbye!", "Decided not to keep running", and then "Finally stopping."
```
thomaseizinger, line 427: I think this comment is outdated because we will never print "Decided not to keep running".

src/send_future.rs
```diff
@@ -32,7 +32,7 @@ pub enum ResolveToHandlerReturn {}
 /// Marker type for resolving the [`SendFuture`] to a [`Receiver`] once the message is queued into the actor's mailbox.
 ///
 /// The [`Receiver`] can be used to await the completion of the handler separately.
-pub enum ResoleToReceiver {}
+pub enum ResolveToReceiver {}
```
thomaseizinger, line 35: The commit history would be cleaner if we would merge changes like this as separate PRs :)

src/context.rs
```diff
@@ -341,12 +343,115 @@ impl<A: Actor> Context<A> {
...
+            fut = unfinished;
+        }
+
+        fut.await
+    }
+
+    /// Handle any incoming messages for the actor while running a given future. This is similar to
+    /// [`Context::join`], but will exit if the actor is stopped, returning the future. Returns
+    /// `Ok` with the result of the future if it was successfully completed, or `Err` with the
+    /// future if the actor was stopped before it could complete.
```
thomaseizinger, line 384: Similar to above, I think a comparison to `futures::future::select` would be appropriate here.

thomaseizinger (approved):
Concept ACK

Borrowing existing ecosystem terminology here is definitely a plus.

src/context.rs
Restioson: Yes - the one difference is that the incoming stream of messages does not usually complete. So, it's like join, but only the user future *must* complete for it to return

src/context.rs
Restioson: > Perhaps Either would be a better return value here? I don't think Result is appropriate because the actor stopping before isn't really a failure case.

That's fair, but I thought that stopping the actor whilst polling in another handler would probably be something similar to failure.

> Additionally, this would then move closer to the futures::select API.

Yea, this makes sense. I think we can change it to either then.

src/send_future.rs
Restioson: True

src/context.rs
Restioson: Good point - I just named them similarly but did not add an explicit comparison

src/context.rs
thomaseizinger: One could argue that an actor stopping is not necessarily a failure case so it can definitely happen.

Restioson: It can happen, sure. What I mean, though, is that if this were semantically equivalent to join(handle_messages, future), then it would not return until the actor stopped. This function, by comparison, will return if the actor doesn't stop if future does.

thomaseizinger: Ah right! Thanks for the clarification.

I guess `join` is still a pretty good name for this although when comparing it to `futures::join` it is not quite accurate.

If we ever get custom event loops, I think this could be removed in favour of the user polling for a new message themselves which then does away with the naming problem :)

Restioson: True. It might still be worth keeping this around in some form for convenience - maybe in an extension crate, though. When custom event loops happen, hopefully the function will only be a few lines, too!

src/send_future.rs
Restioson: I moved this to a commit on master since I think it's probably quite a noncontroversial change, so it doesn't need a whole branch and PR

src/context.rs
Restioson: The user polling on their own could have potential for mis-ordered messages; I will write up how and why in #85 when it's ready for review

Restioson:
Ready for merge?

PR #84, opened 2022-06-13
Scoped task spawning

Restioson:
This is a concept for an API which allows you to spawn tasks that will never outlive the actor. It might be better suited to an extension crate at a later date.

CC @klochowicz

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ For better ergonomics with xtra, try the [spaad](https://crates.io/crates/spaad)
 
 ## Features
 - Safe: there is no unsafe code in xtra.
-- Tiny: xtra is less than 1k loc.
+- Tiny: xtra is less than 2kloc.
 - Lightweight: xtra has few dependencies, most of which are lightweight (except `futures`).
 - Asynchronous and synchronous message handlers.
 - Simple asynchronous message handling interface which allows `async`/`await` syntax even when borrowing `self`.
--- a/examples/scoped_actor_task.rs
+++ b/examples/scoped_actor_task.rs
@@ -44,7 +44,7 @@ async fn main() {
     };
 
     // This task will end as soon as the actor stops
-    let task = tokio::spawn(task.scoped(&addr));
+    let task = tokio::spawn(xtra::scoped(&addr, task));
 
     drop(addr);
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -5,7 +5,7 @@ For better ergonomics with xtra, try the [spaad](https://crates.io/crates/spaad)
 
 ## Features
 - Safe: there is no unsafe code in xtra.
-- Tiny: xtra is less than 1k loc.
+- Tiny: xtra is less than 2kloc.
```
Restioson, line 8: This can be updated again ahead of 0.6 :) Hopefully some convenience methods can be moved out to `xtras` or similar, in order to make the core library smaller, lighter, and more maintainable!

src/lib.rs
```diff
@@ -20,6 +20,9 @@ mod receiver;
 /// This module contains types representing the strength of an address's reference counting, which
 /// influences whether the address will keep the actor alive for as long as it lives.
 pub mod refcount;
+/// This module contains a way to scope a future to the lifetime of an actor, stopping it before it
+/// completes if the actor it is associated with stops too.
+pub mod scoped_task;
```
thomaseizinger, line 25: I think this should be private and re-export the extension trait (if we choose to keep it).

thomaseizinger (commented):
Some thoughts!

examples/scoped_actor_task.rs
```diff
@@ -0,0 +1,51 @@
...
+        .await
+        .expect("Actor should not be dropped");
+
+    let task = async {
+        let _checker = DropChecker;
+        future::pending::<()>().await
+    };
+
+    // This task will end as soon as the actor stops
+    let task = tokio::spawn(task.scoped(&addr));
```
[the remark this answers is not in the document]
Restioson: That could definitely work too

luckysori (approved):
Very nice work! It looks a lot cleaner than all the other approaches we've tried to achieve this on top of `xtra`.

Have you considered writing a test? The example is great, but I suppose you won't be able to know if something is broken unless you run the example and check stdout.

klochowicz (approved):
Looks really cool! thanks for adding the free function, it is indeed easier for discoverability.

Restioson:
Good point, I will add a test and then merge.

PR #100, opened 2022-06-23
Remove `Context::attach_stream` in favor of `StreamExt::forward`

thomaseizinger:
Is it okay to entirely remove the `attach_stream` function from `MessageChannel`? There is no `BoxedSink` unfortunately so I am not sure what to return here.


Restioson:
I think Sink might be implemented for Pin<Box<S>> where S: Sink?

thomaseizinger:
> I think Sink might be implemented for Pin<Box<S\>> where S: Sink?

But what would `S` be in the case of `MessageChannel`?

We can wait for `struct MessageChannel` and use `impl Sink`.

Restioson:
> But what would S be in the case of MessageChannel?

I think it could be `dyn Sink` as the implementation of `AddressSink` should still be obtainable in the implementation of `MessageChannel` for each concrete address type.

> We can wait for struct MessageChannel and use impl Sink.

That is true too

Restioson:
It's also worth noting that the current Sink implementation essentially amounts to `split_receiver()` - it does not wait for the message to be handled before allowing another to be queued.

thomaseizinger:
> > But what would S be in the case of MessageChannel?
> 
> I think it could be `dyn Sink` as the implementation of `AddressSink` should still be obtainable in the implementation of `MessageChannel` for each concrete address type.

Ah yes of course!

src/address.rs
```diff
@@ -202,40 +202,6 @@ impl<A, Rc: RefCounter> Address<A, Rc> {
...
-    ///
-    /// Often, this should be spawned onto an executor to run in the background. **Do not await this
-    /// inside of an actor** - this will cause it to await forever and never receive any messages.
-    ///
-    /// **Note:** if this stream's continuation should prevent the actor from being dropped, this
-    /// method should be called on [`Address`]. Otherwise, it should be called
-    /// on [`WeakAddress`].
-    pub async fn attach_stream<S, M, K>(self, stream: S)
-    where
-        K: Into<KeepRunning> + Send,
```
Restioson, line 218: Could we get rid of KeepRunning now?

Restioson (commented):
Currently this is not quite equivalent to `attach_stream`'s backpressure behaviour, since it doesn't wait for a message to be handled before sending the next. Maybe this is okay, though?

src/address.rs
thomaseizinger: Yes.

thomaseizinger:
> Currently this is not quite equivalent to `attach_stream`'s backpressure behaviour, since it doesn't wait for a message to be handled before sending the next. Maybe this is okay, though?

If we use the `unfold` impl from the other PR, back-pressure would be enforced IIUC. But we will have to fix the tests ..

thomaseizinger:
I can't make the stream cancel test pass. Check the last commit, maybe you have an idea :)

Restioson:
### test_stream_cancel_join`
Forward does not have the same behaviour as `attach_stream`. This behaviour can be regained through `xtra::scoped`. See:
```rust
#[tokio::test]
async fn test_stream_cancel_join() {
    let addr = StreamCancelTester {}.create(None).spawn_global();

    // Attach a stream that blocks forever
    let fwd = stream::pending::<StreamCancelMessage>()
        .map(Ok)
        .forward(addr.downgrade().into_sink());
    let handle = tokio::spawn(xtra::scoped(&addr, fwd));

    // Stop the actor
    drop(addr);

    // Attached stream should return immediately
    assert!(handle.timeout(Duration::from_secs(2)).await.is_some()); // None == timeout
}
```

This makes the test kind of irrelevant, as it is just a `ActorJoinHandle` test. I propose removing it.

### `address_send_exercises_backpressure`

The sink part of this can be removed here as `Send` doesn't multi-stage delivery and waiting for a response

src/address.rs
Restioson: Can it be done in this PR? `attach_stream` was the only reason it was kept this long

thomaseizinger:
Cool! Thanks for the input, I'll update this next week. Busy over the weekend :)

Restioson:
No problem at all, enjoy the weekend :)

thomaseizinger:
@Restioson This is ready for a full review.

Restioson (approved):

PR #101, opened 2022-06-23
Instrumentation by default

Restioson:
This PR changes how xtra's instrumentation works, moving it from an opt-in `.instrumented_child()` to instrumenting every message automatically if the `with-tracing-0_1` feature is enabled. It's slightly less flexible, as now every message will be instrumented as a child, with no option for `follows_from`. There is also no way to opt-out of instrumentation on a per-send basis. I think these are sensible choices, though.

This also expands instrumentation support to separately track the span a message spends in the queue and the span it spends executing. This should allow for finer-grained metrics to be exported via opentelemetry and imported into prometheus for applications using xtra to get more insight into how their actors are performing.

Here is a complex example from a real-world application, [isixhosa.click](https://isixhosa.click), running in a dev environment:

example of how spans are emitted and rendered in jaeger with this change

The fully-qualified actor and message names are listed in each span, so they can be easily filtered by message and actor type.

This should be able to provide [span metrics](https://grafana.com/docs/tempo/latest/server_side_metrics/span_metrics/) to users of Grafana Tempo. Similar solutions may exist for other software.

## TBD
- [x] Test
- [x] Interrupted error
- [x] ~~What log level should these traces be?~~ Set to DEBUG for now
- [x] Broadcast instrumentation
- [x] ~~Find a better name for "xtra message in queue", since it technically includes waiting to be sent into the queue, too - maybe this is ok though?~~ Renamed to "xtra message waiting for actor"


thomaseizinger:
Cool!

With this, we should be able to gather some real world benchmarks in how fast `xtra` actually is :)

klochowicz:
really exciting! 
have you considered adding a separate feature flag for tracing spans / instrumentation instead of reusing `with-tracing-0_1`? There might be some users who are only using tracing for logging, without any opentelemetry or prometheus config. It could also aid discoverability.

Restioson:
>  have you considered adding a separate feature flag for tracing spans / instrumentation instead of reusing with-tracing-0_1? There might be some users who are only using tracing for logging, without any opentelemetry or prometheus config. It could also aid discoverability.

Currently, `with-tracing-0_1` does nothing else except to enable this feature, so even if the user will use `tracing` there is no reason to enable `with-tracing-0_1`, I think

src/envelope.rs
```diff
@@ -114,10 +180,30 @@ impl<A: Handler<M>, M: Send + 'static> MessageEnvelope for NonReturningEnvelope<
...
+            let Instrumentation { parent, .. } = instrumentation;
+            let executing = debug_span!(parent: parent, "actor_message_handler");
+            fut.instrument(executing)
+        };
+
+        Box::pin(fut.map(|_| ()))
     }
 }
 
+// TODO instrument
```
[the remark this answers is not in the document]
Restioson: Currently it's under the TBD list as "Broadcast instrumentation". I am not sure if or how this should be supported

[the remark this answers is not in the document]
thomaseizinger: My bad, didn't read the entire description :)

Restioson:
Ready for merge from my side, pending review :)

src/envelope.rs
```diff
@@ -48,11 +50,46 @@ pub trait MessageEnvelope: Send {
...
+        let parent = debug_span!(
+            parent: Span::current(),
+            "xtra actor request",
+            actor = std::any::type_name::<A>(),
+            message = std::any::type_name::<M>(),
+        );
+
+        // TODO rename: this is technically from send() not in queue alone. Counts waiting to get in
+        // to queue too.
+        let in_queue = debug_span!(
```
klochowicz, line 72: Perhaps this is actionable now by renaming it to `waiting_for_actor`? (with an optional comment stating that it's the time from send() and being it queue together)

klochowicz (approved):
LGTM!

src/envelope.rs
Restioson: Oops - I changed the name but forgot to change the variable and remove TODO :) Thanks!

thomaseizinger:
> > > have you considered adding a separate feature flag for tracing spans / instrumentation instead of reusing with-tracing-0_1? There might be some users who are only using tracing for logging, without any opentelemetry or prometheus config. It could also aid discoverability.
> > 
> > 
> > Currently, `with-tracing-0_1` does nothing else except to enable this feature, so even if the user will use `tracing` there is no reason to enable `with-tracing-0_1`, I think. Discoverability is a good point
> 
> ah, fair enough, if this is the only use of the flag then my previous point was moot. For discoverability, I'd consider renaming the flag to something easier to see the purpose of this feature: `instrumentation`, `tracing-spans` or sth like that.

I'd also be in favor of an `instrumentation` feature.

src/envelope.rs
```diff
@@ -4,6 +4,8 @@ use std::sync::Arc;
 use catty::{Receiver, Sender};
 use futures_core::future::BoxFuture;
 use futures_util::FutureExt;
+#[cfg(feature = "with-tracing-0_1")]
+use tracing::{debug_span, Instrument, Span};
```
thomaseizinger, line 8: You could omit this import if you'd fully-qualify the items.

thomaseizinger: Less `cfg`s are better IMO.

src/envelope.rs
```diff
@@ -48,11 +50,44 @@ pub trait MessageEnvelope: Send {
...
+#[cfg(feature = "with-tracing-0_1")]
+#[derive(Clone)]
+struct Instrumentation {
+    parent: Span,
+    _waiting_for_actor: Span,
+}
+
+#[cfg(feature = "with-tracing-0_1")]
+impl Instrumentation {
+    fn new<A: Actor, M>() -> Self {
```
thomaseizinger, line 62: I believe the `Actor` bound is unnecessary.

src/envelope.rs
```diff
@@ -48,11 +50,44 @@ pub trait MessageEnvelope: Send {
...
+        let parent = debug_span!(
+            parent: Span::current(),
+            "xtra actor request",
```
thomaseizinger, line 65: Is naming of spans with spaces idiomatic? I've not worked with traces much but I could imagine that simple processing on the commandline could be unnecessarily complicated with this having spaces in it.

src/envelope.rs
```diff
@@ -78,12 +115,36 @@ impl<A: Handler<M, Return = R>, M: Send + 'static, R: Send + 'static> MessageEnv
...
+        let fut = {
+            let Instrumentation { parent, .. } = instrumentation;
+            let executing = debug_span!(
+                parent: &parent,
+                "xtra message handler",
+                actor = std::any::type_name::<A>(),
+                message = std::any::type_name::<M>(),
+            );
+            fut.instrument(executing)
+        };
```
thomaseizinger, line 145: If we would delegate this to `Instrumentation`, like:

```rust
let fut = instrumentation.apply(fut);
```

Then we could also do this unconditionally and simply have a no-op impl for when the feature is not enabled.

thomaseizinger: This would also reduce code duplication between the various envelopes.

thomaseizinger (approved):
Concept ACK

A few ideas on how to reduce the complexity coming from the many `cfg`s!

src/envelope.rs
Restioson: The other option is bounding with `'static` I think - it's used by the `type_name` in this function

src/envelope.rs
```diff
@@ -48,11 +50,44 @@ pub trait MessageEnvelope: Send {
...
+        }
+    }
+}
+
 /// An envelope that returns a result from a message. Constructed by the `AddressExt::do_send` method.
 pub struct ReturningEnvelope<A, M, R> {
     message: M,
     result_sender: Sender<R>,
     phantom: PhantomData<fn() -> A>,
+    #[cfg(feature = "with-tracing-0_1")]
```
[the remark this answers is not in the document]
Restioson: Good point

src/envelope.rs
thomaseizinger: I'd prefer that.

src/envelope.rs
Restioson: I've only used Jaeger as a trace explorer, where it does not matter. Maybe we should change it to _s to be safe

src/envelope.rs
Restioson: Nevermind, `type_name` doesn't need any bound :smile:

Restioson:
I've changed the feature to `instrumentation` and addressed the other comments :)

Cargo.toml
```diff
@@ -55,11 +55,11 @@ async-std = { version = "1.0", features = ["attributes"] }
 [features]
 default = ["timing"]
 timing = ["futures-timer"]
+instrumentation = ["tracing"]
```
thomaseizinger, line 58:
```suggestion
instrumentation = ["dep:tracing"]
```

This will hide the `tracing` feature which will otherwise show up [here](https://docs.rs/crate/xtra/latest/features) but not do anything.

Can be done in a follow-up PR though because we should be doing the same thing for all the others too!

BREAKING-CHANGES.md
```diff
@@ -14,6 +14,7 @@
 - Previously, `stop_all` would immediately disconnect the address. However, `stop_self` done on every actor would actually
   not do this in one case - if there were a free-floating (not executing an actor event loop) Context. This change brings
   `stop_all` in line with `stop_self`.
+- `InstrumentedExt` was removed. All messages are now instrumented automatically when `with-tracing-0_1` is enabled.
```
thomaseizinger, line 17:
```suggestion
- `InstrumentedExt` was removed. All messages are now instrumented automatically when `instrumentation` is enabled.
```

src/envelope.rs
```diff
@@ -114,7 +182,17 @@ impl<A: Handler<M>, M: Send + 'static> MessageEnvelope for NonReturningEnvelope<
...
+        let Self {
+            message,
+            instrumentation,
+            ..
+        } = *self;
+
+        Box::pin(
+            instrumentation
+                .apply::<A, M, _>(act.handle(message, ctx))
+                .map(|_| ()),
```
thomaseizinger, line 194: Nit: We could remove this if we would restrict `Handler` to a `Return` of `()`.

thomaseizinger (approved):
Nice!

Changelog should be changed in this PR, everything else is optional.

src/envelope.rs
Restioson: True, but as of this change it is still used with any `Return`, so this can wait for a followup I think

Cargo.toml
Restioson: Good point :) Let's move this to a followup as you suggested

PR #102, opened 2022-06-23
Introduce `MessageChannel` struct

### LINKED ISSUE #68: Discoverability for MessageChannel (recipients)
Actix has support for [`Recipient`](https://actix.rs/actix/actix/prelude/struct.Recipient.html), an alternative to `Address` where the actor type is hidden, leaving only the message itself. This is extremely useful for building large APIs across multiple crates where being able to name actors causes dependency cycles.

How feasible would it be for `xtra` to support this?

thomaseizinger:
Fixes #68.

cc @zesterer What do you think?


Restioson:
This is really nice! It's more flexible and also less code, which I didn't expect :)

Restioson (approved):
This looks great, thanks so much :)

PR #103, opened 2022-06-23
Replace `AddressSink` with `sink::unfold` and `impl Trait`

thomaseizinger:
Thoughts?

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -10,11 +10,11 @@ use std::pin::Pin;
 use std::task::{Context, Poll};
 
 use event_listener::EventListener;
-use futures_core::{FusedFuture, Stream};
+use futures_core::Stream;
 use futures_sink::Sink;
 use futures_util::{future, FutureExt, StreamExt};
 
-use crate::envelope::{NonReturningEnvelope, ReturningEnvelope};
+use crate::envelope::ReturningEnvelope;
 use crate::inbox::{PriorityMessageToOne, SentMessage};
 use crate::refcount::{Either, RefCounter, Strong, Weak};
 use crate::send_future::ResolveToHandlerReturn;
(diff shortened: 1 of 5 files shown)


src/address.rs
```diff
@@ -325,47 +329,3 @@ impl<A, Rc: RefCounter> Hash for Address<A, Rc> {
...
     }
 }
-
-/// A sink which can be used to send messages to an actor. These messages will have default priority
-/// and be handled in send order.
-pub struct AddressSink<A, Rc: RefCounter>(inbox::SendFuture<A, Rc>);
-
-impl<A, Rc: RefCounter> AddressSink<A, Rc> {
-    /// Return a clone of the underlying [`Address`] which this sink sends to.
-    pub fn address(&self) -> Address<A, Rc> {
```
Restioson, line 335: The main thing if we implement this change is losing this method (and the nameability of the type)

thomaseizinger:
> Strange test failure... fails only on all features?

No tests are executed without the tokio feature!

The test fails because a single poll is no longer enough to send the item. I believe that is because the new impl waits for the handler too!

I think I may be able to make it pass if I use `split_receiver`.

src/address.rs
thomaseizinger: Also, Rust will hopefully soon support impl-trait-type-aliases! :)

Restioson:
> I believe that is because the new impl waits for the handler too!

Ah, I see! 🤔 In that case, maybe sink can get a different test, since it wouldn't make sense anymore to test it like this

tests/basic.rs
```diff
@@ -753,18 +754,6 @@ async fn address_send_exercises_backpressure() {
...
-    let mut sink = address.into_sink();
-    let _ = sink
-        .send(PrintHello("world"))
-        .now_or_never()
-        .expect("be able to queue another message because the mailbox is empty again");
-
-    assert!(
-        sink.send(PrintHello("world")).now_or_never().is_none(),
-        "Fail to queue 2nd message because mailbox is full"
-    );
```
thomaseizinger, line 766: I've deleted this test because the implementation is so simply, I don't think we need to test it and the existence of the method is tested via the example.

Restioson: Agreed

Restioson (approved):
LGTM! 

For a later PR, we could see if supporting priority and broadcast through sinks (either with different kinds of sinks or with a message wrapper) would be desireable, but this has the same features as master so I'm okay with merging this now

PR #105, opened 2022-06-24
Make split_receiver() resolve to Result<Receiver>

Restioson:
This change lets users figure out from `split_receiver()` futures awaited once whether the message was actually delivered or not. There are maybe other ways to accomplish this from an API perspective, though.

### DIFF
--- a/src/receiver.rs
+++ b/src/receiver.rs
@@ -17,12 +17,6 @@ pub struct Receiver<R> {
 }
 
 impl<R> Receiver<R> {
-    pub(crate) fn disconnected() -> Self {
-        Self {
-            inner: Inner::Disconnected,
-        }
-    }
-
     pub(crate) fn new(receiver: catty::Receiver<R>) -> Self {
         Self {
             inner: Inner::Receiving(receiver.map_err(|_| Disconnected)),
@@ -31,7 +25,6 @@ impl<R> Receiver<R> {
 }
 
 enum Inner<R> {
-    Disconnected,
     Receiving(MapErr<catty::Receiver<R>, fn(catty::Disconnected) -> Disconnected>),
     Done,
 }
@@ -43,10 +36,6 @@ impl<R> Future for Receiver<R> {
         let this = self.get_mut();
 
         match mem::replace(&mut this.inner, Inner::Done) {
-            Inner::Disconnected => {
-                this.inner = Inner::Done;
-                Poll::Ready(Err(Disconnected))
-            }
             Inner::Receiving(mut rx) => match rx.poll_unpin(cx) {
                 Poll::Ready(item) => {
                     this.inner = Inner::Done;
--- a/src/send_future.rs
+++ b/src/send_future.rs
@@ -128,7 +128,7 @@ impl<A, R, Rc: RefCounter> SetPriority for NameableSending<A, R, Rc> {
 }
 
 impl<A, R, Rc: RefCounter> Future for NameableSending<A, R, Rc> {
-    type Output = Receiver<R>;
+    type Output = Result<Receiver<R>, Disconnected>;
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         let this = self.get_mut();
@@ -136,8 +136,8 @@ impl<A, R, Rc: RefCounter> Future for NameableSending<A, R, Rc> {
         let result = futures_util::ready!(this.inner.poll_unpin(cx));
 
         match result {
-            Ok(()) => Poll::Ready(this.receiver.take().expect("polled after completion")),
-            Err(_) => Poll::Ready(Receiver::disconnected()),
+            Ok(()) => Poll::Ready(Ok(this.receiver.take().expect("polled after completion"))),
+            Err(_) => Poll::Ready(Err(Disconnected)),
         }
     }
 }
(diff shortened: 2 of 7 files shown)


src/receiver.rs
```diff
@@ -31,7 +25,6 @@ impl<R> Receiver<R> {
 }
 
 enum Inner<R> {
-    Disconnected,
     Receiving(MapErr<catty::Receiver<R>, fn(catty::Disconnected) -> Disconnected>),
     Done,
 }
```
thomaseizinger, line 30: If this has only two states now, is it even worth having this enum?

I think we can hold the `MapErr` as a regular field now and just poll that.

thomaseizinger: Together with that, I think this code should move into `send_future`, esp. now that we have multiple `Receiver`s flying around :D

thomaseizinger (approved):
This makes sense to me, one suggestion!

src/receiver.rs
Restioson: > I think we can hold the MapErr as a regular field now and just poll that.

I think that returns Poll::Pending after completed instead of panicking - we can move away from this if that's acceptable

thomaseizinger: > > I think we can hold the MapErr as a regular field now and just poll that.
> 
> I think that returns Poll::Pending after completed instead of panicking - we can move away from the inner enum if that's acceptable

I think that is fine. No one should expect any specific behaviour from futures after their completion.

Restioson: Done now

src/send_future.rs
```diff
@@ -193,10 +213,11 @@ where
 
         match mem::replace(&mut this.inner, SendFutureInner::Done) {
             SendFutureInner::Sending(mut send_fut) => match send_fut.poll_unpin(ctx) {
```
thomaseizinger, line 215: I think you could use `?` here.


```suggestion
            SendFutureInner::Sending(mut send_fut) => match send_fut.poll_unpin(ctx)? {
```

thomaseizinger (approved):
I think this is a good change although adding another `Result` to the API is a bit of an ergonomics hit.

Restioson:
> adding another Result to the API is a bit of an ergonomics hit.

That's true in the case of `split_receiver().await.unwrap().await`, but I feel like this usecase is far less common than `let _ = addr.send(...).split_receiver().await.unwrap()`

thomaseizinger:
> > adding another Result to the API is a bit of an ergonomics hit.
> 
> That's true in the case of `split_receiver().await.unwrap().await`, but I feel like this usecase is far less common than `let _ = addr.send(...).split_receiver().await.unwrap()`
> 

Uhm, I am not sure. I was more thinking of what I would do to properly handle this error.

Could we funnel that error into the `Result` on the `Receiver`?

Restioson:
> Uhm, I am not sure. I was more thinking of what I would do to properly handle this error.

If you `await` the receiver, you get the `Err` immediately anyway. However, if you *don't* `await` it, there is no way to check whether the message was received except with `.now_or_never()` or some other way to poll once, which seems like a crutch for API inflexibility IMO

> Could we funnel that error into the Result on the Receiver?

How do you mean?

thomaseizinger:
> > Uhm, I am not sure. I was more thinking of what I would do to properly handle this error.
> 
> If you `await` the receiver, you get the `Err` immediately anyway. However, if you *don't* `await` it, there is no way to check whether the message was received except with `.now_or_never()` or some other way to poll once, which seems like a crutch for API inflexibility IMO

Fair.

> > Could we funnel that error into the Result on the Receiver?
> 
> How do you mean?

Someone needs to await the receiver. Not doing that I'd assume to be a bad design.

Now by using `split_receiver`, a user opts into asynchronous behaviour with all its quirks and edge-cases. For example, it shouldn't really be the case that the _sender_ of a message wants both: async processing AND feedback on how the processing ended. If you want the latter, use sync, if you want the former, design your system such that knowing the outcome is not necessary.

Hence, knowing whether the queuing of an async message worked should be good enough at the receiver end. If the sender cares, they shouldn't use async.

Restioson:
This behaviour is currently required to reimplement the `SendAsyncSafe` trait from itchysats. I don't fully understand what the use case of this trait is, though, so I will try understand that first :sweat_smile:

thomaseizinger:
> This behaviour is currently required to reimplement the `SendAsyncSafe` trait from itchysats. I don't fully understand what the use case of this trait is, though, so I will try understand that first :sweat_smile: 

From my memory, that behaviour is already covered because you can do `split_receiver` and still access the return value.

For unit returns, that receiver can be spawned into a runtime and be forgotten about :)

Restioson:
> that behaviour is already covered because you can do split_receiver and still access the return value.

The behaviour of `SendAsyncSafe` currently is to send the message and then immediately return once it's been delivered into the mailbox. If this fails since the actor is disconnected, then it returns early. Maybe this could be replaced with `is_connected()` after sending, though... This has slightly different semantics of course but it might be good enough

thomaseizinger:
> > that behaviour is already covered because you can do split_receiver and still access the return value.
> 
> The behaviour of `SendAsyncSafe` currently is to send the message and then immediately return once it's been delivered into the mailbox. If this fails since the actor is disconnected, then it returns early. Maybe this could be replaced with `is_connected()` after sending, though... This has slightly different semantics of course but it might be good enough

I can't speak for it now but the main design intent in the past was to just not drop return values when using async handling and we implemented that with the APIs available on xtra back then :)

cc @luckysori who also worked on these traits.

Is there a change in how back-pressure is handled with this PR?

All in all, I am okay with this change. Perhaps the added semantics is worth the API-ergonomics hit.

Restioson:
> Is there a change in how back-pressure is handled with this PR?

I do not believe so, no

thomaseizinger:
This has been shipped as part of #147.

PR #107, opened 2022-06-24
Remove the 'static bound on A of Address

Restioson:
This removes the `'static` bound on `A` of `Address`. It also moves some inline bounds into `where` clauses, and changes the `PhantomData` to better reflect how `A` it is actually related to the envelope types.


Restioson:
Uh oh, weird diff.

PR #110, opened 2022-06-24
Remove the 'static bound on A of Address (take 2)

Restioson:
Same as #107 but cleaner history


thomaseizinger (approved):
Nice!

Can we remove `'static' in other places as a result?

Restioson:
I'll take a look

Restioson:
Nothing else in the public API except A: Actor in ActorManager

thomaseizinger (approved):

PR #112, opened 2022-06-28
Add TickFuture and TickFuture::get_or_create

Restioson:
This PR adds a function which returns the span of the inner message handler, so that it can be recorded and logged to by a custom event loop.

## Motivating example

```rust
pub async fn run(mut self, mut actor: A) -> A::Stop {
    loop {
        let msg = self.ctx.next_message().await;

        let (span, fut) = self.ctx.tick_instrumented(msg, &mut actor);
        match timeout(self.timeout, fut).await {
            Ok(ControlFlow::Continue(())) => (),
            Ok(ControlFlow::Break(())) => break actor.stopped().await,
            Err(_elapsed) => {
                if let Some(span) = span {
                    let _entered = span.enter();
                    let timeout_seconds = self.timeout.as_secs();
                    span.record("interrupted", &"timed_out");
                    tracing::warn!(%timeout_seconds, "Handler execution timed out");
                }
            }
        }
    }
}
```

Without this change, there would be no way to log that the message handler timed out to any child span of `xtra_actor_request`.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -494,7 +494,7 @@ impl<'a, A> Future for TickFuture<'a, A> {
     fn poll(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Self::Output> {
         match mem::replace(&mut self.state, TickState::Done) {
             TickState::Running { mut fut, phantom } => {
-                match self.span.0.in_scope(|| fut.poll_unpin(cx)) {
+                match self.span.in_scope(|| fut.poll_unpin(cx)) {
                     Poll::Ready(flow) => Poll::Ready(flow),
                     Poll::Pending => {
                         self.state = TickState::Running { fut, phantom };
--- a/src/context.rs
+++ b/src/context.rs
@@ -414,7 +414,6 @@ pub struct TickFuture<'a, A> {
     span: HandlerSpan,
 }
 
-#[cfg(feature = "instrumentation")]
 impl<'a, A> TickFuture<'a, A> {
     /// Return the handler's [`tracing::Span`](https://docs.rs/tracing/latest/tracing/struct.Span.html),
     /// creating it if it has not already been created. This can be used to log messages into the
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,15 +1,17 @@
 use std::future::Future;
 use std::ops::ControlFlow;
 use std::pin::Pin;
-use std::task;
 use std::task::Poll;
+use std::{mem, task};
 
+use futures_core::future::BoxFuture;
 use futures_core::FusedFuture;
 use futures_util::future::{self, Either};
 use futures_util::FutureExt;
 #[cfg(feature = "timing")]
 use {futures_timer::Delay, std::time::Duration};
 
+use crate::envelope::HandlerSpan;
 use crate::inbox::rx::{ReceiveFuture as InboxReceiveFuture, RxStrong};
 use crate::inbox::ActorMessage;
 use crate::{inbox, Actor, Address, Error, Handler, WeakAddress};
@@ -131,47 +133,8 @@ impl<A: Actor> Context<A> {
     }
 
     /// Handle one message and return whether to exit from the manage loop or not.
-    pub async fn tick(&mut self, msg: Message<A>, actor: &mut A) -> ControlFlow<()> {
-        match msg.0 {
-            ActorMessage::ToOneActor(msg) => msg.handle(actor, self).0.await,
-            ActorMessage::ToAllActors(msg) => msg.handle(actor, self).0.await,
-            ActorMessage::Shutdown => {
-                self.running = false;
-                ControlFlow::Break(())
-            }
-        }
-    }
-
-    /// Handle one message, returning the span of the actor message handler. This is very similar
-    /// to [`Context::tick`].
-    ///
-    /// This span is pre-filled with an `interrupted` field which can be `Span::record`ed to in the
-    /// case of the handler being cancelled. This function returns the span and the future to handle
-    /// the message as two parts of a tuple. The span will not be present if the message is a shut
-    /// down message, as there would be no message to handle.
-    #[cfg(feature = "instrumentation")]
-    pub fn tick_instrumented<'a>(
-        &'a mut self,
-        msg: Message<A>,
-        actor: &'a mut A,
-    ) -> (
-        Option<tracing::Span>,
-        impl Future<Output = ControlFlow<()>> + 'a,
-    ) {
-        match msg.0 {
-            ActorMessage::ToOneActor(msg) => {
-                let (fut, span) = msg.handle(actor, self);
-                (Some(span.0), Either::Left(fut))
-            }
-            ActorMessage::ToAllActors(msg) => {
-                let (fut, span) = msg.handle(actor, self);
-                (Some(span.0), Either::Left(fut))
-            }
-            ActorMessage::Shutdown => {
-                self.running = false;
-                (None, Either::Right(future::ready(ControlFlow::Break(()))))
-            }
-        }
+    pub fn tick<'a>(&'a mut self, msg: Message<A>, actor: &'a mut A) -> TickFuture<'a, A> {
+        TickFuture::new(msg, self, actor)
     }
 
     /// Yields to the manager to handle one message, returning the actor should be shut down or not.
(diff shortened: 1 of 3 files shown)


thomaseizinger:
Can we build that as a feature into the `Drop` of the `NextMessage` future?

Sorry for the bare bones review, I haven't got much time on my hands :)

thomaseizinger:
Ah `tick` doesn't return a named future but is just an `async fn`. Can we make it a named future and apply the above idea?

I think there is also a possibility to move the entire `tick` function onto `Message` IF we make `Message`  have a `&mut` to `Context` inside (would need a lifetime from `Context` and have `next_message` take `&mut self`).

Perhaps that helps too?

If possible, I would really like to avoid having duplication like this but favour orthogonal APIs.

Restioson:
> Would a function on NextMessage that returns the Span also suffice?

This span isn't actually constructed until the message is handled. I'm a bit tired now so I'll think about this tomorrow more :)

thomaseizinger:
> > Would a function on NextMessage that returns the Span also suffice?
> 
> This span isn't actually constructed until the message is handled. I'm a bit tired now so I'll think about this tomorrow more :)

Yeah I realized that too which is why I wrote that 2nd comment 😅

We'd have to make `tick` return a named future.

Restioson:
I've given what you suggested a shot!

src/context.rs
```diff
@@ -403,3 +403,107 @@ impl<A> FusedFuture for ReceiveFuture<A> {
...
+                    Box::pin(async move {
+                        ctx.running = false;
+                        ControlFlow::Break(())
+                    }),
+                    None,
+                )
+            }
+        };
+
+        (fut, Some(span))
```
thomaseizinger, line 477: If we would do this straight in the ctor, the span would be directly available and instead of the `with_span` function, we could have:

```rust
fn span(&self) -> Span
```

I'd make it panic if the future is complete already.

Is there a reason to access the span before polling the future? Otherwise we could also do it in the first poll and return `None` before or panic.

Thoughts?

src/context.rs
```diff
@@ -403,3 +403,107 @@ impl<A> FusedFuture for ReceiveFuture<A> {
...
+pub struct TickFuture<'a, A> {
+    state: TickState<'a, A>,
+}
+
+#[cfg(feature = "instrumentation")]
+impl<'a, A> TickFuture<'a, A> {
+    /// Return the handler's [`tracing::Span`](https://docs.rs/tracing/latest/tracing/struct.Span.html)
+    /// and the future, if the message is not a shut down. This can be used to log messages into the
+    /// span when required, such as if it is cancelled later due to a timeout.
+    pub fn with_span(self) -> (Option<tracing::Span>, TickFuture<'a, A>) {
```
thomaseizinger, line 416: Can we even call this on a pinned future? Once it has been polled, it will be pinned.

src/envelope.rs
```diff
@@ -57,6 +58,9 @@ struct Instrumentation {
     _waiting_for_actor: tracing::Span,
 }
 
+#[derive(Clone)]
+pub struct HandlerSpan(#[cfg(feature = "instrumentation")] pub tracing::Span);
```
thomaseizinger, line 62: Do we need this encapsulation with the new `TickFuture`?

src/context.rs
```diff
@@ -403,3 +403,107 @@ impl<A> FusedFuture for ReceiveFuture<A> {
...
+                self.poll_unpin(cx)
+            }
+            TickState::Running { mut fut, span } => match fut.poll_unpin(cx) {
+                Poll::Ready(flow) => Poll::Ready(flow),
+                Poll::Pending => {
+                    self.state = TickState::Running { fut, span };
+                    Poll::Pending
+                }
+            },
+            TickState::Done => Poll::Pending,
```
thomaseizinger, line 506: Panic instead?

thomaseizinger (commented):
Looks much better!

A few questions. Plus I think we should have an example for this :)

src/context.rs
Restioson: That's true, I do not believe so - example might make the intended use clearer though so I'll add one

src/envelope.rs
Restioson: It's still used elsewhere in `envelope` iirc

src/context.rs
Restioson: Sure

src/context.rs
thomaseizinger: > > Is there a reason to access the span before polling the future? Otherwise we could also do it in the first poll and return None before or panic.
> 
> Do you mean to make the future resolve to `(Span, Fut)`?

No, I was thinking of having an _accessor_ to get the span out of the `Future`. In other words, instead of consuming self, it would just be &self. Something like:

```rust
match futures::select(sleep(1), ctx.tick(msg, &mut actor)) {
    Either::Left(((), unfinished_tick)) => {
        let _entered = unfinished_tick.span().enter();

        tracing::warn!(...);
    }
}
```

Restioson: This could be added in addition to the current function, I think. I don't think `fn span(&self) -> &Span` is possible, since `Shutdown` messages don't have a handle span, considering they don't invoke a handler, and are not currently instrumented.

thomaseizinger: > This could be added in addition to the current function, I think.

Why not instead? I find the `with` API a bit awkward to use because we are not passing anything to it but it returns a tuple.

Restioson: If instead, it would need to be &mut self, unless it created the span upon future creation. The downside of that, though, is that it would start early.

The span creation is going to be leaked anyway by its nature, since when it starts is measurable.

thomaseizinger: > If instead, it would need to be &mut self, unless it created the span upon future creation. The downside of that, though, is that it would start early.

We could do it on the first poll and have it return `None` before if we have to have `Option` in the signature anyway because of `Shutdown`.

Restioson: > We could do it on the first poll and have it return None before if we have to have Option in the signature anyway because of Shutdown.

This is essentially the behaviour now, but it you can create it early. Otherwise, you need to poll it and then still access the future which is usually quite unergonomic

thomaseizinger: > > We could do it on the first poll and have it return None before if we have to have Option in the signature anyway because of Shutdown.
> 
> This is essentially the behaviour now, but it you can create it early. Otherwise, you need to poll it and then still access the future which is usually quite unergonomic

Is it? I am not sure I agree. Did you see my timeout example based on `select`?

I find it not clean that I have to create a `Span` outside of a code branch (before the poll), if it is only ever used within one code branch (the error case). As a rule of thumb, I tend to give variables a lifetime and scope that is as short/narrow as possible. 

Another view could be to say that `TickFuture` has a lifetime to `Context` anyway so you can't really do anything other than almost immediately poll it and thus, creating `Span` in the ctor doesn't falsify measurements much and always by the same amount for a given codebase whilst giving users the freedom to access the span whenever they want.

Restioson: > Is it? I am not sure I agree. Did you see my timeout example based on select?

I think this may be the exception rather than the rule, given that most methods of polling a future do not give access to the future after its first poll - it consumes it.

> Another view could be to say that TickFuture has a lifetime to Context anyway so you can't really do anything other than almost immediately poll it and thus, creating Span in the ctor doesn't falsify measurements much and always by the same amount for a given codebase whilst giving users the freedom to access the span whenever they want.

That's fair

thomaseizinger: > > Is it? I am not sure I agree. Did you see my timeout example based on select?
> 
> I think this may be the exception rather than the rule, given that most methods of polling a future do not give access to the future after its first poll - it consumes it.

That is true too! I guess using `select` is more of a hack then?

What if we put the `Span` into a `RefCell<Option>`? We should be able to guarantee that no two mutable references exist at the same time so we should be able to create the span either on first poll or on first access through a `fn span(&self)` function.

Restioson: Another option would be to make it `get_or_create_span(&mut self)`?

thomaseizinger: > Another option would be to make it `get_or_create_span(&mut self)`?

As long as there is no issue with pinning here, I think that works.

Should we go with `span` as function name and hide the "get or create" in the docs to make it less noisy?

Restioson: > As long as there is no issue with pinning here, I think that works.

The future is `Unpin`, so I think it should work?

> Should we go with span as function name and hide the "get or create" in the docs to make it less noisy?

I think `get_or_create` is better as it's more explicit and contains the explanation for why &mut self and not &self in its name, as well as it's intuitively showing the span would start from that call.

thomaseizinger: Fine with me :)

Restioson: Done. I have a followup which fixes an issue where the Span would get created too early (I elaborated on this in xtra-community)

src/context.rs
```diff
@@ -404,3 +408,123 @@ impl<A> FusedFuture for ReceiveFuture<A> {
...
+    ///          let _entered = span.enter();
+    ///          span.record("interrupted", &"timed_out");
+    ///          tracing::warn!(timeout_seconds = 1, "Handler execution timed out");
+    ///      }
+    ///  }
+    /// # } })
+    /// ```
+    ///
+    #[cfg(feature = "instrumentation")]
+    pub fn get_or_create_span(&mut self) -> &tracing::Span {
```
thomaseizinger, line 454: Would it be more ergonomic to not return a reference here?

thomaseizinger: We are immediately cloning it in the example and the reference keeps the mutable borrow alive which means the future cannot be polled during that time. I don't see how a user would not immediately clone the span to get rid of the reference.

thomaseizinger (commented):
One question, otherwise LGTM

src/context.rs
Restioson: > I don't see how a user would not immediately clone the span to get rid of the reference.

There are some functions that you can use on the span which only take &self, such as [`record`](https://docs.rs/tracing/latest/tracing/span/struct.Span.html#method.record), which is why I kept it like this with an explicit clone in most cases

thomaseizinger: Fair enough.

thomaseizinger (approved):

thomaseizinger:
Commit message or PR title should be fixed to reflect the new API :)

PR #115, opened 2022-07-08
Instrument attach_stream

Restioson:
Adds a log and span to the function to make it a bit clearer while debugging. Otherwise, it just shows up as an `xtra_actor_request` like any other. This is a bit more detailed.

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -215,8 +215,19 @@ impl<A, Rc: RefCounter> Address<A, Rc> {
                 future::select(&mut stream.next(), &mut stopped).await
             {
                 let res = self.send(m); // Bound to make it Sync
-                if matches!(res.await.map(Into::into), Ok(KeepRunning::Yes)) {
-                    continue;
+
+                #[cfg(feature = "instrumentation")]
+                let (res, span) =  {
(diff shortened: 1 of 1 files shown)


thomaseizinger:
Is this worth doing with https://github.com/Restioson/xtra/pull/100 on the way?

Restioson:
How close do you think we are to getting #100 in?

Restioson:
Okay, we can leave this for now then I think! I think this should be reimplementable with `with` + `forward`

Restioson:
Actually, I don't think that would be possible, but maybe when instrumentation is required a custom loop is better anyway

thomaseizinger:
I would have expected that instrumenting the `Forward` future would work but haven't tested it.

Restioson:
I think that would create one span for the entire lifetime of the stream vs one per forward operation

thomaseizinger:
> I think that would create one span for the entire lifetime of the stream vs one per forward operation

But the existence of this plus the regular request span tells you what is going on, no? Or is the span only visible once it ends? That could be a problem for never ending "forward"s.

Restioson:
Apparently `Instrument` works on streams too!

PR #118, opened 2022-07-11
Move `Context::stop_all` to `Address::stop_all`

thomaseizinger:
Stopping all actors is equivalent to sending a `Stop` message to
each actor. Sending messages is done on an `Address`. Thus stopping
all actors should be done on an `Address` too.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -17,9 +17,10 @@
 - `Context::attach_stream` was removed in favor of composing `Stream::forward` and `Address::into_sink`.
 - `KeepRunning` was removed as `Context::attach_stream` was its last usage.
 - `InstrumentedExt` was removed. All messages are now instrumented automatically when `instrumentation` is enabled.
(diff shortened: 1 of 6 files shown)


Restioson:
Given the current conclusion in #119, do you still think we should move this, or no?

Restioson:
> I will need to think about it a bit but I still find this to be somewhat more natural on Address yes.

The main thing IMO is that it's a bit odd to have `stop` be only on `Context`, but `stop_all` only on `Address`. I understand why this is, but it's a bit strange. The other thing is that it changes the encapsulation of the actor a bit. Previously, the only way to shut down an actor from the outside would be to explicitly write a message to do so, or to drop all addresses. Now, there is an additional way, which means that the actor's stopping behaviour is no longer uniquely controlled by the actor itself anymore. I would say that dropping all addresses is an exception to this rule as this is not generally a part of control flow but more about liveness.

thomaseizinger:
I agree for it to remain on `Context`.

PR #120, opened 2022-07-12
Create spans late

Restioson:
This PR changes the instrumentation to create spans only after first send poll. This allows us to instrument the futures from send with `.send(...).instrument(parent)` properly, having the send span come under the `parent` span.

### DIFF
--- a/src/envelope.rs
+++ b/src/envelope.rs
@@ -1,5 +1,6 @@
 use std::future::Future;
 use std::marker::PhantomData;
+use std::ops::ControlFlow;
 use std::sync::Arc;
 
 use catty::{Receiver, Sender};
@@ -46,7 +47,7 @@ pub trait MessageEnvelope: Send {
         self: Box<Self>,
         act: &'a mut Self::Actor,
         ctx: &'a mut Context<Self::Actor>,
-    ) -> BoxFuture<'a, ()>;
+    ) -> (BoxFuture<'a, ControlFlow<()>>, HandlerSpan);
 }
 
 #[derive(Clone)]
@@ -55,6 +56,31 @@ struct Instrumentation {
     parent: tracing::Span,
     #[cfg(feature = "instrumentation")]
     _waiting_for_actor: tracing::Span,
+}
+
+#[derive(Clone)]
+pub struct HandlerSpan(#[cfg(feature = "instrumentation")] pub tracing::Span);
+
+impl HandlerSpan {
+    pub fn in_scope<R>(&self, f: impl FnOnce() -> R) -> R {
+        #[cfg(feature = "instrumentation")]
+        let r = self.0.in_scope(f);
+
+        #[cfg(not(feature = "instrumentation"))]
+        let r = f();
+
+        r
+    }
+
+    fn disabled() -> HandlerSpan {
+        #[cfg(feature = "instrumentation")]
+        let span = HandlerSpan(tracing::Span::none());
+
+        #[cfg(not(feature = "instrumentation"))]
+        let span = HandlerSpan();
+
+        span
+    }
 }
 
 impl Instrumentation {
@@ -85,7 +111,7 @@ impl Instrumentation {
         Instrumentation {}
     }
 
-    fn apply<A, M, F>(self, fut: F) -> impl Future<Output = F::Output>
+    fn apply<A, M, F>(self, fut: F) -> (impl Future<Output = F::Output>, HandlerSpan)
     where
         F: Future,
     {
@@ -96,13 +122,17 @@ impl Instrumentation {
                 "xtra_message_handler",
                 actor = std::any::type_name::<A>(),
                 message = std::any::type_name::<M>(),
+                interrupted = tracing::field::Empty,
             );
 
-            tracing::Instrument::instrument(fut, executing)
+            (
+                tracing::Instrument::instrument(fut, executing.clone()),
+                HandlerSpan(executing),
+            )
         }
 
         #[cfg(not(feature = "instrumentation"))]
-        fut
+        (fut, HandlerSpan())
     }
 }
 
@@ -140,7 +170,7 @@ where
         self: Box<Self>,
         act: &'a mut Self::Actor,
         ctx: &'a mut Context<Self::Actor>,
-    ) -> BoxFuture<'a, ()> {
+    ) -> (BoxFuture<'a, ControlFlow<()>>, HandlerSpan) {
         let Self {
             message,
             result_sender,
@@ -148,14 +178,17 @@ where
             ..
         } = *self;
 
-        Box::pin(
-            instrumentation
-                .apply::<A, M, _>(act.handle(message, ctx))
-                .map(move |r| {
-                    // We don't actually care if the receiver is listening
-                    let _ = result_sender.send(r);
-                }),
-        )
+        let fut = async move { (act.handle(message, ctx).await, ctx.flow()) };
+
+        let (fut, span) = instrumentation.apply::<A, M, _>(fut);
+
+        let fut = Box::pin(fut.map(move |(r, flow)| {
+            // We don't actually care if the receiver is listening
+            let _ = result_sender.send(r);
+            flow
+        }));
+
+        (fut, span)
     }
 }
 
@@ -167,7 +200,7 @@ pub trait BroadcastEnvelope: HasPriority + Send + Sync {
         self: Arc<Self>,
         act: &'a mut Self::Actor,
         ctx: &'a mut Context<Self::Actor>,
-    ) -> BoxFuture<'a, ()>;
+    ) -> (BoxFuture<'a, ControlFlow<()>>, HandlerSpan);
 }
 
 pub struct BroadcastEnvelopeConcrete<A, M> {
@@ -199,10 +232,15 @@ where
         self: Arc<Self>,
         act: &'a mut Self::Actor,
         ctx: &'a mut Context<Self::Actor>,
-    ) -> BoxFuture<'a, ()> {
+    ) -> (BoxFuture<'a, ControlFlow<()>>, HandlerSpan) {
         let (msg, instrumentation) = (self.message.clone(), self.instrumentation.clone());
         drop(self); // Drop ASAP to end the message waiting for actor span
-        Box::pin(instrumentation.apply::<A, M, _>(act.handle(msg, ctx)))
+        let fut = async move {
+            act.handle(msg, ctx).await;
+            ctx.flow()
+        };
+        let (fut, span) = instrumentation.apply::<A, M, _>(fut);
+        (Box::pin(fut), span)
     }
 }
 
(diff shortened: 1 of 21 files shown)


src/envelope.rs
```diff
@@ -72,53 +56,61 @@ impl HandlerSpan {
...
 
-    fn disabled() -> HandlerSpan {
+    fn none() -> Span {
         #[cfg(feature = "instrumentation")]
-        let span = HandlerSpan(tracing::Span::none());
+        let span = Span(tracing::Span::none());
 
         #[cfg(not(feature = "instrumentation"))]
-        let span = HandlerSpan();
+        let span = Span();
```
thomaseizinger, line 64: It would make the review easier to not have this rename in here, or at least in a commit at the very front of the PR :)

src/inbox.rs
```diff
@@ -89,7 +89,7 @@ impl<A> ChanInner<A> {
         // If len < cap after popping this message, try fulfill at most one waiting sender
         if capacity.map_or(false, |cap| cap == self.ordered_queue.len()) {
             match self.try_fulfill_sender(MessageType::Ordered) {
-                Some(SentMessage::ToOneActor(msg)) => self.ordered_queue.push_back(msg.val),
+                Some(MessageKind::ToOneActor(msg)) => self.ordered_queue.push_back(msg.val),
```
thomaseizinger, line 92: Same with this rename!

src/envelope.rs
```diff
@@ -21,47 +21,31 @@ pub trait MessageEnvelope: Send {
     /// The type of actor that this envelope carries a message for
     type Actor;
 
-    /// Handle the message inside of the box by calling the relevant `AsyncHandler::handle` or
-    /// `Handler::handle` method, returning its result over a return channel if applicable. The
-    /// reason that this returns a future is so that we can propagate any `Handler` responder
-    /// futures upwards and `.await` on them in the manager loop. This also takes `Box<Self>` as the
+    /// Starts the instrumentation of this message request. This will create the request span.
+    fn start_span(&mut self, msg_name: &'static str);
```
thomaseizinger, line 25: Is this really the best way we can implement this?

From reading the PR description "create span on first poll", I did not expect something like this but a change to `poll` of `SendFuture`. I haven't tried implementing this myself so perhaps this is impossible but I am finding the indirection here quite confusing.

Also not really happy with the amount of conditional code that we are growing here.

thomaseizinger (commented):
Some initial thoughts.

I think we should support what is outlined in the PR description but I find it quite hard to follow the implementation atm.

I'll play around with this myself and see what I can come up with!

src/inbox.rs
Restioson: This was done since SentMessage became a struct with more stuff inside of it

src/envelope.rs
Restioson: > From reading the PR description "create span on first poll", I did not expect something like this but a change to poll of SendFuture. I haven't tried implementing this myself so perhaps this is impossible but I am finding the indirection here quite confusing.

I thought about this but wasn't entirely sure what the best design would be for this so I prototyped it this way (part of the work was getting the semantics right about when the span should actually be created).

One early concept I had was storing the message in `SendFuture` in a new `New` variant of the inner type. However, this refers to the actor type, so I wasn't sure whether this would work with the `sending_erased` version.

> Also not really happy with the amount of conditional code that we are growing here.

Are you referring to feature-conditional/cfg code or branches?

thomaseizinger:
I have a branch locally that moves some stuff around in the hope that features like this will be easier to build.

I am happy to merge this for now IF we can add a test that ensures my upcoming refactoring is not gonna break anything 😁

thomaseizinger:
> > I am happy to merge this for now IF we can add a test that ensures my upcoming refactoring is not gonna break anything
> 
> Hmm, I wonder how we can write tests for tracing. I don't really have experience with testing for side-effects of the program like logs! I can look into this tomorrow.

I wrote a custom subscriber here at some point to test log output: https://github.com/comit-network/xmr-btc-swap/blob/master/swap/src/tracing_ext.rs

src/envelope.rs
thomaseizinger: It would be nice if we could flatten out the layers in such a way that we can start the instrumentation at a point where we still now the message and actor types but are still outside the `SendFuture`.

> > From reading the PR description "create span on first poll", I did not expect something like this but a change to poll of SendFuture. I haven't tried implementing this myself so perhaps this is impossible but I am finding the indirection here quite confusing.
> 
> I thought about this but wasn't entirely sure what the best design would be for this so I prototyped it this way (part of the work was getting the semantics right about when the span should actually be created).
> 
> One early concept I had was storing the message in `SendFuture` in a new `New` variant of the inner type. However, this refers to the actor type, so I wasn't sure whether this would work with the `sending_erased` version.
> 
> > Also not really happy with the amount of conditional code that we are growing here.
> 
> Are you referring to feature-conditional/cfg code or branches?

Feature-conditional code.

Restioson:
I've added a test now

tests/tracing.rs
```diff
@@ -0,0 +1,143 @@
...
+    type Return = ();
+
+    async fn handle(&mut self, message: Hello, _ctx: &mut Context<Self>) {
+        tracing::info!("Hi {}", message.0)
+    }
+}
+
+#[tokio::test]
+async fn assert_send_is_child_of_span() {
+    let mock_writer = MockWriter::new(&GLOBAL_BUF);
```
thomaseizinger, line 130: Do we need to use a global buffer here? Looks trivial to use a function-local buffer which would allow us to write more of these tests that can be run in parallel, plus we can drop the lazy_static dependency.

tests/tracing.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    let (addr, mut ctx) = Context::<Tracer>::new(None);
+    let _ = addr
+        .send(Hello("hi"))
+        .split_receiver()
+        .instrument(tracing::info_span!("Test span"))
+        .await;
+    ctx.yield_once(&mut Tracer).await;
+
+    logs_assert(|lines: &[&str]| lines[0].contains("Test span:xtra_actor_request"));
```
thomaseizinger, line 142: It would be nice to assert the entire log line here :)

We can disable timestamps to make sure the test is not dependent on time.

tests/tracing.rs
Restioson: Good point!

tests/tracing.rs
Restioson: We could do this, but I'm not sure if the format is semver guaranteed?

thomaseizinger: I would expect them to make an effort to not break it and even if it does, fixing the test seems like a small effort for the benefit of a more readable assertion (that for example contains the actual log message too).

tests/tracing.rs
thomaseizinger: If we wrap the buffer in a new type, we can hide the details of converting the bytes to a lines iterator.

Restioson: Is there much point when this will mostly just be used through `with_logs` anyway?

Restioson:
Uh oh, committed with someone else as author :sweat_smile: let me take this opportunity to squash some stuff up

tests/tracing.rs
```diff
@@ -0,0 +1,146 @@
...
+        .split_receiver()
+        .instrument(tracing::info_span!("Test span"))
+        .await;
+    ctx.yield_once(&mut Tracer).await;
+
+    with_logs(&buf, |lines: &[&str]| {
+        assert_eq!(
+            lines,
+            [
+                r#" INFO Test span:xtra_actor_request{actor="tracing::Tracer" tracing::Hello}:xtra_message_handler{actor="tracing::Tracer" tracing::Hello}: tracing: Hi hi"#
```
thomaseizinger, line 142: I think we can actually see here that using `message` as a span variable name has weird side effects. Should maybe call it `msg_type` or something?

tests/tracing.rs
```diff
@@ -0,0 +1,146 @@
...
+    let buf = Arc::new(Mutex::new(vec![]));
+    let mock_writer = MockWriter::new(buf.clone());
+    let subscriber = get_subscriber(mock_writer, "tracing=trace,xtra=trace");
+    let _g = tracing::dispatcher::set_default(&subscriber);
+
+    let (addr, mut ctx) = Context::<Tracer>::new(None);
+    let _ = addr
+        .send(Hello("hi"))
+        .split_receiver()
+        .instrument(tracing::info_span!("Test span"))
```
thomaseizinger, line 134: Can we use a span name here without spaces? That would make it easier to spot in the assertion.

tests/tracing.rs
thomaseizinger, line 142: I think the last "tracing" here is the test's name right? Can we rename that too? It is a bit confusing to read so many things named tracing :)

Restioson: That's a good point, actually. It seems to work okay with Jaeger (and Grafana Tempo), so I never noticed this

tests/tracing.rs
thomaseizinger: Maybe call it `user_span` or something?

tests/tracing.rs
Restioson: No, the last `tracing` is the log target, which defaults to the module

Restioson: Speaking of this, I was thinking of removing the tags from everything but `xtra_actor_request`, otherwise they repeat. Should this and the rename go in a separate PR?

Restioson: We could just override it, I think

tests/tracing.rs
```diff
@@ -0,0 +1,147 @@
...
+async fn assert_send_is_child_of_span() {
+    let buf = Arc::new(Mutex::new(vec![]));
+    let mock_writer = MockWriter::new(buf.clone());
+    let subscriber = get_subscriber(mock_writer, "tracing=trace,xtra=trace");
+    let _g = tracing::dispatcher::set_default(&subscriber);
+
+    let (addr, mut ctx) = Context::<Tracer>::new(None);
+    let _ = addr
+        .send(Hello("hi"))
+        .split_receiver()
```
thomaseizinger, line 134: I find the usage of `split_receiver` a bit noisy here. Why not just use `spawn_global` here?

Using `split_receiver` here seems like it matters for the test.

tests/tracing.rs
```diff
@@ -0,0 +1,147 @@
...
+#[tokio::test]
+async fn assert_send_is_child_of_span() {
+    let buf = Arc::new(Mutex::new(vec![]));
+    let mock_writer = MockWriter::new(buf.clone());
+    let subscriber = get_subscriber(mock_writer, "tracing=trace,xtra=trace");
+    let _g = tracing::dispatcher::set_default(&subscriber);
+
+    let (addr, mut ctx) = Context::<Tracer>::new(None);
+    let _ = addr
+        .send(Hello("hi"))
```
thomaseizinger, line 133: Nit:


```suggestion
        .send(Hello("world"))
```

tests/tracing.rs
thomaseizinger: Right, I was meant to say it is the test _file_'s name.

tests/tracing.rs
Restioson: > I find the usage of split_receiver a bit noisy here. Why not just use spawn_global here?

The default subscriber is thread local, I believe, so we need to handle the message on the same thread. It might also change with tokio task - I'm not sure. To make sure everything definitely worked, I wanted to do everything on the same task and thread.

tests/tracing.rs
Restioson: :laughing:

tests/tracing.rs
thomaseizinger: > Speaking of this, I was thinking of removing the tags from everything but `xtra_actor_request`, otherwise they repeat. Should this and the rename go in a separate PR?

I'd prefer a 2nd PR yeah :)

tests/tracing.rs
thomaseizinger: > > I find the usage of split_receiver a bit noisy here. Why not just use spawn_global here?
> 
> The default subscriber is thread local, I believe, so we need to handle the message on the same thread. It might also change with tokio task - I'm not sure. To make sure everything definitely worked, I wanted to do everything on the same task and thread.

I believe the default tokio test executor is single-threaded and if not, you can customise it via the `tokio::test` attribute.

tests/tracing.rs
Restioson: I've renamed it to `instrumentation`

tests/tracing.rs
thomaseizinger: Yep, it is single-threaded: https://docs.rs/tokio/latest/tokio/attr.test.html#using-default

Restioson: I've changed it now

Restioson:
Merged with master now

PR #122, opened 2022-07-14
Move channel logic into `Chan`

thomaseizinger:
As part of trying to come up with an alternative to https://github.com/Restioson/xtra/pull/120, I started to move a bit of code around and ended up with this patch set!

Each commit in its own builds and has passing tests and I was careful to try and not alter any behaviour. In other words, this is a pure refactoring.

The goal was as follows:

- Have `Chan` house the core channel implementation with descriptive function names for key operations
- Have `Sender` and `Receiver` be thin layers on top that combine `Chan` with reference counting and provide a `Future`-based APIs for sending and receiving

I can extract some of these commits into individual PRs if you want. The diff will be quite hard to follow once it is squashed.

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -12,8 +12,6 @@ use event_listener::EventListener;
 use futures_sink::Sink;
 use futures_util::FutureExt;
 
-use crate::envelope::ReturningEnvelope;
-use crate::inbox::{tx, PriorityMessageToOne, SentMessage};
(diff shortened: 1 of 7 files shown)


Restioson (approved):

thomaseizinger:
I just noticed absolute irony. This PR starts with a commit that removes `Sender::send` and ends with a commit that adds it again. I guess I've gone full circle 😂

PR #124, opened 2022-07-14
Remove outdated changelog entry

thomaseizinger:
These traits no longer exist within latest master.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -2,7 +2,6 @@
 
 ## 0.6.0
 
-- Sealed `RefCounter` and `MessageChannel` traits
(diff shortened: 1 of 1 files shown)


Restioson:
Doesn't RefCounter still exist?

thomaseizinger:
> Doesn't RefCounter still exist? 

Technically no, there are two of them now 😁

I guess because the 2nd one didn't exist before, we don't need to mention it so in regards to breaking changes, it is only the one that got sealed.

Restioson:
Well, RxRefCounter isn't public, but TxRefCounter still is (and, if I'm not wrong, is still sealed). TxRefCounter is also exposed under the name RefCounter I think

Restioson:
Thanks!

PR #130, opened 2022-07-16
Re-design feature set

### LINKED ISSUE #109: Re-work exposed features
I am using this issue to create two proposals:

1. We should use the new `dep:` syntax for features to hide the implicit features created by optional crates. I think this part is non-controversial.
2. Rename the features we expose to their original crate name. The argument would be that we are currently not supporting multiple versions of the same dependency and thus the `with-xyz-1_0` convention is unnecessarily complex.

Together with this issue, the README should be updated too.

thomaseizinger:
Fixes #109.

This also makes use of the new `dep:` syntax and thus hides the implicit features generated by optional dependencies.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
         toolchain: [
           stable,
           beta,
-          1.56.0 # MSRV
+          1.60.0 # MSRV
         ]
         args: [
           --all-features,
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,6 +110,7 @@ members = ["examples/basic_wasm_bindgen"]
 
 [package.metadata.docs.rs]
 features = ["async_std", "smol", "tokio", "wasm_bindgen"]
+rustdoc-args = ["--cfg", "docsrs"]
 
 [[bench]]
 name = "throughput"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ documentation = "https://docs.rs/xtra"
 readme = "README.md"
 keywords = ["async", "actor", "futures", "xtra", "async-await"]
 categories = ["asynchronous", "concurrency"]
-rust-version = "1.56.0"
+rust-version = "1.60.0"
 
 [dependencies]
 async-trait = "0.1.56"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,16 +25,9 @@ spin = { version = "0.9.3", default-features = false, features = ["spin_mutex"] 
 # Feature `timing`
 futures-timer = { version = "3.0", optional = true, default-features = false }
 
-# Feature `with-async_std-1`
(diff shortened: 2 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ jobs:
         toolchain: [
           stable,
           beta,
-          1.56.0 # MSRV
+          1.60.0 # MSRV
```
Restioson, line 15: Out of interest, what is the reason for this bump?

Cargo.toml
```diff
@@ -54,69 +47,70 @@ async-std = { version = "1.0", features = ["attributes"] }
...
-instrumentation = ["tracing"]
-with-async_std-1 = ["async-std"]
-with-smol-1 = ["smol"]
-with-tokio-1 = ["tokio"]
-with-wasm_bindgen-0_2 = ["wasm-bindgen", "wasm-bindgen-futures"]
+instrumentation = ["dep:tracing"]
+async_std = ["dep:async-std"]
+smol = ["dep:smol"]
+tokio = ["dep:tokio"]
+wasm_bindgen = ["dep:wasm-bindgen", "dep:wasm-bindgen-futures"]
```
Restioson, line 54: This is nice!

README.md
```diff
@@ -73,22 +73,22 @@ Too verbose? Check out the [spaad](https://crates.io/crates/spaad) sister-crate!
...
 ## Okay, sounds great! How do I use it?
 Check out the [docs](https://docs.rs/xtra) and the [examples](https://github.com/Restioson/xtra/blob/master/examples)
-to get started! Enabling the `with-tokio-1`, `with-async_std-1`, `with-smol-1`, or `with-wasm_bindgen-0_2` features
+to get started! Enabling the `tokio`, `async_std`, `smol`, or `wasm_bindgen` features
 is recommended in order to enable some  convenience methods (such as `Actor::spawn`). Which you enable will depend on
 which executor you want to use (check out their docs to learn more about each). If you have any questions, feel free to
 [open an issue](https://github.com/Restioson/xtra/issues/new) or message me on the [Rust discord](https://bit.ly/rust-community).
 
-Keep in mind that `xtra` has a MSRV of 1.56.0.
+Keep in mind that `xtra` has a MSRV of 1.60.0.
```
Restioson, line 81: This should be updated in Cargo.toml

Restioson (commented):
Thanks! Other than the one thing I mentioned this is ready for merge IMO

.github/workflows/ci.yml
thomaseizinger: The `dep:` syntax! It was only stabilised in 1.60.0.

thomaseizinger:
Conflicts resolved and comments addressed.

.github/workflows/ci.yml
Restioson: Ah, I see

Restioson:
Thanks!

PR #131, opened 2022-07-16
Remove legacy doc attributes from `Context::join`

thomaseizinger:
These annotations generate a warning when compiling the docs with
a nightly compiler. We also do now have a separate example in the
documentation of this function so embedding the other example is
not necessary.


Restioson:
Thanks!

Issue #126, opened 2022-07-14
Re-focus public API on `Address` & `Mailbox` aka short-lived `Context`


thomaseizinger:
As part of doing https://github.com/Restioson/xtra/pull/122, I started to form the following vision:

- `Chan` should be the core channel implementation of the library, #122 gets us most of the way there
- `Sender` and `Receiver` only really add an opinionated interfaces on top of it, plus reference counting
- Neither `Sender` nor `Receiver` are publicly exposed, meaning I think they don't actually have to exist
- `Sender` can be completely inlined into `Address`, removing an entire layer of indirection
- `Receiver` can be promoted to a type called `Mailbox` that could be publicly exposed
- `Context` would be something that is constructed only temporarily for the invocation of a handler (I think we discussed this idea at some point)

Putting all of this together, we could realise an API like this:

```rust
let (mailbox, address) = Mailbox::bounded(5);

tokio::spawn(xtra::run(mailbox, MyActor::new()));
```

With `run` being implemented more or less like what we have today.

Restioson:
I think this makes sense!

PR #133, opened 2022-07-16
Make `Context` short-lived

thomaseizinger:
This is the final PR towards #126.

## Open questions:

- Instead of passing `Mailbox` to `started`, should we pass `Context` too but use a lifecycle type parameter to control, whether users can call `stop` or not?

## Open tasks

- [x] Add changelog entry
- [x] Adapt README to new API


examples/interleaved_messages.rs
```diff
@@ -22,8 +22,9 @@ impl Handler<Hello> for ActorA {
 
     async fn handle(&mut self, _: Hello, ctx: &mut Context<Self>) {
         println!("ActorA: Hello");
-        let fut = self.actor_b.send(Hello);
-        ctx.join(self, fut).await.unwrap();
+        xtra::join(ctx.mailbox(), self, self.actor_b.send(Hello))
```
thomaseizinger, line 25: Initially, I thought we would get away with not exposing the `Mailbox` through the `Context` but if we don't, these usecases become impossible.

src/lib.rs
```diff
@@ -147,7 +154,7 @@ pub trait Actor: 'static + Send + Sized {
 
     /// Called as soon as the actor has been started.
     #[allow(unused_variables)]
-    async fn started(&mut self, ctx: &mut Context<Self>) {}
+    async fn started(&mut self, mailbox: &mut Mailbox<Self>) {}
```
thomaseizinger, line 157: Not entirely sure about this change. It feels a bit weird to pass the `Mailbox` but the upside is that people can no longer call `stop` in here which simplifies the event loop.

src/lib.rs
```diff
@@ -211,3 +218,189 @@ impl fmt::Display for Error {
...
+pub async fn run<A>(mut mailbox: Mailbox<A>, mut actor: A) -> A::Stop
+where
+    A: Actor,
+{
+    actor.started(&mut mailbox).await;
+
+    while let ControlFlow::Continue(()) = yield_once(&mut mailbox, &mut actor).await {}
+
+    return actor.stopped().await;
+}
```
thomaseizinger, line 235: I mean, this is pretty cool.

examples/crude_bench.rs
```diff
@@ -127,10 +127,10 @@ async fn do_parallel_address_benchmark<R>(
 ) where
     SendFuture<(), NameableSending<Counter, ()>, R>: Future,
 {
-    let (addr, ctx) = Context::new(None);
+    let (addr, mailbox) = Mailbox::new(None);
```
thomaseizinger, line 130: I quite like the way this reads because `Address` and `Mailbox` are counterparts so it makes sense that they are instantiated together.

thomaseizinger: This is also much more inline with various channel APIs.

src/lib.rs
Restioson: I feel like this should also be possible on master, actually 🤔

thomaseizinger: What type are you going to pass into `Actor::started`? You'd have to do something like #76 I think.

thomaseizinger:
One thing that would be cool to do on top of this is to remove the `Sender` and `Receiver` types so we have one less layer between the internal channel implementation and the public types.

src/lib.rs
Restioson: I mean that the `loop` can be replaced by the `while let` - I think the if is still necessary unfortunately

thomaseizinger: > I mean that the `loop` can be replaced by the `while let` - I think the if is still necessary unfortunately

Ah yes, the `while let` is also possible on master :)

src/lib.rs
```diff
@@ -211,3 +220,189 @@ impl fmt::Display for Error {
...
+///
+/// # #[cfg(feature = "with-smol-1")]
+/// # smol::block_on(async {
+/// let addr = MyActor.create(None).spawn(&mut xtra::spawn::Smol::Global);
+/// assert!(addr.is_connected());
+/// assert_eq!(addr.send(Selecting).await, Ok(true)); // Assert that the select did end early
+/// # })
+///
+/// ```
+pub async fn select<A, F, R>(mailbox: &mut Mailbox<A>, actor: &mut A, mut fut: F) -> Either<R, F>
```
Restioson, line 318: Is there a reason to have this as a free function rather than on `Mailbox` itself?

src/lib.rs
Restioson: Would it be possible to just pass the address in?

src/lib.rs
thomaseizinger: I figured it doesn't really fit on `Mailbox`. Why not on `Actor`? Why not an extension trait for `Future`s? I don't see how any of these types are "privileged" in where this code should live so a free function felt like the right thing to do.

Restioson: I guess. Personally I feel like it fits on mailbox since you are selecting between mailbox messages and the given future. That being said, though, you are also selecting on handling those messages

thomaseizinger: The 2nd reason is a taste thing. Rust is not an OOP language and the main reason for functions to live _on_ a type is for information hiding purposes where the function needs to access internal state. This is not the case here so it can just be a free function.

src/lib.rs
thomaseizinger: If you just pass the address in, then the user can't call functions like `xtra::join` or `xtra::select` in `started` which is a regression in terms of features.

I think it could be convinced that it is the right thing to do.

src/lib.rs
thomaseizinger: I just drafted this up as quickly as possible, the order of the parameters should probably be thought about a bit more :)

thomaseizinger: What is your latest position on this? Would you like this function to be on `Mailbox` or can I move forward on the free function based design?

src/lib.rs
Restioson: It could be quite interesting to have Context have a second argument, representing the life cycle stage. So, handle would take Context<A, Running> and started would take Context<A, Starting>. Stop would only exist for Context<A, Running>. On the other hand, not being able to call stop in started is technically a feature regression too 🤷‍♂️

thomaseizinger: I'd have `started` return `Result<(), Self::Stop>` so you can still stop the actor in `started`, just not via calling `stop`.

A custom event loop would get a warning from the `must_use` on `Result` so they are hinted towards handling that.

We can certainly do the life-cycle type parameter thing, I'll have a play to see how it turns out! I think the advantage would be that all functions just take a `Context`, including `started`.

It will make custom event loops a bit more verbose because users have to construct a `Context` just for being able to call `started`.

src/lib.rs
Restioson: I think I mentioned this elsewhere but I agree that a free function based design makes sense here

src/lib.rs
Restioson: > A custom event loop would get a warning from the must_use on Result so they are hinted towards handling that.

This is indeed a nice advantage! It might be worth the asymmetry not being able to use Context::stop here like elsewhere (I wonder if this is worth worrying about. I might be a bit confused here as a user)

thomaseizinger:
This is currently blocked by these PRs:

- https://github.com/Restioson/xtra/pull/158
- https://github.com/Restioson/xtra/pull/159

thomaseizinger:
> This is currently blocked by these PRs:
> 
>     * [Introduce generic `ChanPtr` #158](https://github.com/Restioson/xtra/pull/158)
> 
>     * [Introduce top-level `recv_future` module #159](https://github.com/Restioson/xtra/pull/159)

I've squash-merged both PRs into here so that I could resolve the merge conflicts already.

Restioson:
Is this PR ready?

thomaseizinger:
> Is this PR ready?

Yes!

thomaseizinger:
> Is this PR ready?

It probably needs a changelog entry before we can merge it though. Plus the README needs to be adjusted too!

src/mailbox.rs
```diff
@@ -1,30 +1,57 @@
...
 }
 
 impl<A> Mailbox<A> {
-    pub fn new(inner: chan::Ptr<A, Rx>) -> Self {
-        Mailbox {
-            broadcast_mailbox: inner.new_broadcast_mailbox(),
-            inner,
-        }
+    /// Creates a new [`Mailbox`] with the given capacity.
+    pub fn bounded(capacity: usize) -> (Address<A>, Mailbox<A>) {
```
thomaseizinger, line 18: This API is :heart_eyes:

src/mailbox.rs
```diff
@@ -1,30 +1,57 @@
...
 
-use crate::chan::{self, BroadcastQueue, Rx, TxStrong, TxWeak};
+use crate::chan::{self, BroadcastQueue, Rx};
 use crate::recv_future::ReceiveFuture;
+use crate::{Address, WeakAddress};
 
+/// A [`Mailbox`] is the counter-part to an [`Address`].
+///
+/// Messages sent into an [`Address`] will be received in an actor's [`Mailbox`].
+/// Think of [`Address`] and [`Mailbox`] as an MPMC channel.
```
thomaseizinger, line 10: Also this discovery makes me super happy! I think this should be really easy to grasp for users!

tests/basic.rs
```diff
@@ -912,69 +925,6 @@ fn test_addr_cmp_hash_eq() {
...
-    async fn started(&mut self, ctx: &mut Context<Self>) {
-        if self.stop_all {
-            ctx.stop_all();
-        }
-    }
-
-    async fn stopped(self) {
-        println!("Actor {} stopped", self.number);
-    }
-}
```
thomaseizinger, line 976: This is deleted because the new `Actor` trait no longer allows calling `stop` in `started`.

src/lib.rs
```diff
@@ -210,3 +223,191 @@ impl fmt::Display for Error {
...
+pub fn tick<'a, A>(
+    message: Message<A>,
+    actor: &'a mut A,
+    mailbox: &'a mut Mailbox<A>,
+) -> TickFuture<'a, A>
+where
+    A: Actor,
+{
+    TickFuture::new(message.0, actor, mailbox)
+}
```
thomaseizinger, line 267: I am not yet super happy about this. Do you think we can merge this into `ReceiveFuture` somehow?

thomaseizinger: Perhaps not into `ReceiveFuture` itself but the indirection here and also the `Span` handling is rather confusing IMO. It would be nice if users would just need to call `Mailbox::next`, get a message and pass it to the actor.

src/lib.rs
```diff
@@ -174,7 +185,9 @@ pub trait Actor: 'static + Send + Sized {
 
     /// Called as soon as the actor has been started.
     #[allow(unused_variables)]
-    async fn started(&mut self, ctx: &mut Context<Self>) {}
+    async fn started(&mut self, mailbox: &mut Mailbox<Self>) -> Result<(), Self::Stop> {
```
Restioson, line 188: This might introduce a strange pattern of 
```rust
async fn started(..) -> Result<(), Result<(), Error>>
```

thomaseizinger: I've never had `Result` as the `Stop` value of an actor. Seems weird to me?

thomaseizinger: If we just make it return `Stop`, we'd have to trait bound it to somehow construct a default value so it is an optional function.

Restioson: What I would mean is that if you wanted to exit early, you might want to return an error. However, you might want to always return some Ok value from stopped, e.g some statistics. Therefore, to satisfy this api, stop would be a result.

src/lib.rs
Restioson: > and pass it to the actor

Isn't this essentially what `tick` is doing?

thomaseizinger: Yeah, what I find unnecessary is that we need to provide a function for that. It seems like such a simple operation that the user should be able to perform (in a custom event loop f.e.) without calling into a free function of xtra.

I believe the reason for this is that `TickFuture` currently encapsulates the matching on the message variants etc. Perhaps we can move this somewhere else.

Restioson: Does tick not also encapsulate adding the actor to the mix? The next message future just takes the mailbox, whereas tick also borrows the actor.

thomaseizinger: Yep. But really the actor is only needed to then dispatch the message received from the mailbox. it is two operations in fact. I have something locally that I can hopefully put up soon :)

src/lib.rs
thomaseizinger: Hmm, should we use `ControlFlow` here instead of `Result` to not have the "error" notion in it?

thomaseizinger: Another way to look at this is that a double result is not actually bad. See https://sled.rs/errors.html.

src/lib.rs
Restioson: When you say "it", do you mean receive + tick?

src/lib.rs
Restioson: Fair enough! I'm OK with keeping it like this, then.

src/lib.rs
thomaseizinger: Yeah!

To get the message you only need the `Mailbox` and to dispatch it you only need the actor.

thomaseizinger: In any case, I don't think it needs to block this PR!

Restioson (approved):

Issue #121, opened 2022-07-12
Document goals and non-goals of xtra


thomaseizinger:
In https://github.com/Restioson/xtra/issues/119 and other places, we have discussions that ultimately come down to optimising for different goals.

I am opening this issue to propose and discuss a manifest that states goals and non-goals and rates properties amongst each other.

From the current issue description, xtra wants to be safe, small and performant. Taking inspiration from the agile manifest, we can thus perhaps say that we value:

- **Safety over performance**

Example: We would rather not use unsafe code to achieve a more performant implementation.

- **Small API surface over performance**

Example: We would rather not add a "special" function to the public API that makes certain use cases more efficient if the use case can already be achieved with a different API.

- **Small API surface over "batteries-included"**

Example: We would rather not include a convenience feature that can already be expressed with existing public APIs.

- **Convenient APIs over ordering-guarentees**

Example: async-await style APIs are convenient to use but it is hard to provide ordering guarantees once a task is spawned into an executor. Ordering guarantees more or less imply `poll` style APIs down to user-handlers but those are less convenient to use.

Same as with the Agile manifesto, this list doesn't mean that we don't optimise for the items on the right but when in conflict with the left, we will favor the left.

Restioson:
I definitely agree that we should clarify these, and I agree with most of the points!

> Convenient APIs over ordering-guarentees

Could you give an example of when this tradeoff would apply to xtra? I don't quite understand this one.

thomaseizinger:
> I definitely agree that we should clarify these, and I agree with most of the points!

Good to hear!

> > Convenient APIs over ordering-guarentees
> 
> Could you give an example of when this tradeoff would apply to xtra? I don't quite understand this one.

I think this one may need a bit of fleshing out still. I have three thoughts on it:

1. I think the choice of letting users write their handlers with async-await means we give up some opportunities for the user to strictly control ordering.

Every `await` call represents a yield point and thus, the exact order in which a fleet of actors will process something is hard to predict. If I use a multi-threaded executor and send a message to two actors - both with `split_receiver` - then I don't have any guarantees, which one will process the message first.

2. Ordering guarantees and asynchronous processing are fundamentally incompatible.

`split_receiver` is essentially saying "please process this message in an asynchronous manner relative to the current task". So even without an async-await style API, by opting into async processing the user accepts a reordering of operations and has to design their system to deal with that.

3. An actor system is generally not a good fit for systems which rely on strict ordering of operations.

[...]

PR #138, opened 2022-07-16
Remove `ReceiveFuture::cancel` in favor of cancellation-safety docs

### LINKED ISSUE #134: `ReceiveFuture::cancel` is too nuanced to be part of the public API
Based on the docs and how it is used in `Context::select` I think I understand what this API is for. However, putting myself into the shoes of an outside user, I think this API is too nuanced to be exposed / exist and when encountering it, it is hard to "do the right thing" (i.e. choose between `Drop` and `cancel`) without understanding the internals of `xtra`.

If I understand correctly, this goes back to https://github.com/Restioson/xtra/issues/94 and the fact that binary heaps are not insertion order stable. If we had only a fixed set of priorities (as suggested in https://github.com/Restioson/xtra/issues/92), we could just re-insert the message at the front without having messages perceived as out of order, correct?

I think we should work towards removing the `ReceiveFuture::cancel` API.

Thoughts?

thomaseizinger:
Fixes #134.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -275,9 +275,8 @@ impl<A: Actor> Context<A> {
                 let mut next_msg = self.next_message();
                 match future::select(fut, &mut next_msg).await {
                     Either::Left((future_res, _)) => {
-                        if let Some(msg) = next_msg.cancel() {
-                            self.tick(msg, actor).await;
-                        }
+                        let message = next_msg.await; // If don't await the future, message ordering might be compromised.
+                        self.tick(message, actor).await;
 
                         return Either::Left(future_res);
                     }
@@ -315,22 +314,17 @@ impl<A> Clone for Context<A> {
 pub struct Message<A>(pub(crate) ActorMessage<A>);
 
 /// A future which will resolve to the next message to be handled by the actor.
+///
+/// # Cancellation safety
+///
+/// This future is cancellation-safe in that no messages will ever be lost, even if this future is
+/// dropped half-way through. However, reinserting the message into the mailbox may mess with the
+/// ordering of messages and they may be handled by the actor out of order.
+///
+/// If the order in which your actors process messages is not important to you, you can consider this
+/// future to be fully cancellation-safe.
 #[must_use = "Futures do nothing unless polled"]
 pub struct ReceiveFuture<A>(InboxReceiveFuture<A, RxStrong>);
-
-impl<A> ReceiveFuture<A> {
-    /// Cancel the receiving future, returning a message if it had been fulfilled with one, but had
-    /// not yet been polled after wakeup. Future calls to `Future::poll` will return `Poll::Pending`,
-    /// and `FusedFuture::is_terminated` will return `true`.
-    ///
-    /// This is important to do over `Drop`, as with `Drop` messages may be sent back into the
-    /// channel and could be observed as received out of order, if multiple receives are concurrent
-    /// on one thread.
-    #[must_use = "If dropped, messages could be lost"]
-    pub fn cancel(&mut self) -> Option<Message<A>> {
-        self.0.cancel().map(Message)
-    }
-}
 
 impl<A> Future for ReceiveFuture<A> {
     type Output = Message<A>;
--- a/src/context.rs
+++ b/src/context.rs
@@ -326,6 +326,10 @@ pub struct Message<A>(pub(crate) ActorMessage<A>);
 ///
 /// If the order in which your actors process messages is not important to you, you can consider this
 /// future to be fully cancellation-safe.
+///
+/// If you wish to maintain message ordering, you can use [`FutureExt::now_or_never`] to do a final
+/// poll on the future. [`ReceiveFuture`] is guaranteed to complete in a single poll if it has
+/// remaining work to do.
 #[must_use = "Futures do nothing unless polled"]
 pub struct ReceiveFuture<A>(InboxReceiveFuture<A, RxStrong>);
 
--- a/src/context.rs
+++ b/src/context.rs
@@ -275,8 +275,11 @@ impl<A: Actor> Context<A> {
                 let mut next_msg = self.next_message();
                 match future::select(fut, &mut next_msg).await {
                     Either::Left((future_res, _)) => {
-                        let message = next_msg.await; // If don't await the future, message ordering might be compromised.
-                        self.tick(message, actor).await;
+                        // Check if `ReceiveFuture` was in the process of handling a message.
+                        // If yes, dispatch it, otherwise continue.
+                        if let Some(message) = next_msg.now_or_never() {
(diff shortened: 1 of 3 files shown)


src/context.rs
```diff
@@ -275,9 +275,8 @@ impl<A: Actor> Context<A> {
                 let mut next_msg = self.next_message();
                 match future::select(fut, &mut next_msg).await {
                     Either::Left((future_res, _)) => {
-                        if let Some(msg) = next_msg.cancel() {
-                            self.tick(msg, actor).await;
-                        }
+                        let message = next_msg.await; // If don't await the future, message ordering might be compromised.
```
Restioson, line 278: Consider the following:

- Future is `futures::future::ok(())`
- There is one address which is sleeping for 10 minutes and will send a message later

In this case, this `select` would wait the full 10 minutes

thomaseizinger: Right, in this case, `cancel` would have returned `None` right?

Restioson: Yea. On the other hand, I think `now_or_never` could race, on the other hand...

thomaseizinger: I see the problem, so this can only be solved if:

- We accept the potential re-ordering of messages
- We somehow make sure we can re-insert messages into the mailbox without re-ordering

thomaseizinger: > Yea. On the other hand, I think `now_or_never` could race, on the other hand...

Oh, that is an interesting idea! Whether or not this works might depend on how many `poll`s we need to dispatch? Do you have any idea on how we could test this?

thomaseizinger: I managed to write a test! Let me know what you think.

tests/basic.rs
```diff
@@ -1082,3 +1082,13 @@ fn no_sender_returns_disconnected() {
...
 }
+
+#[tokio::test]
+async fn receive_future_can_dispatch_in_one_poll() {
+    let (addr, ctx) = Context::<Greeter>::new(None);
+
+    let _ = addr.send(Hello("world")).split_receiver().await;
+    let receive_future = ctx.next_message();
+
+    assert!(receive_future.now_or_never().is_some())
```
Restioson, line 1093: I think the race here would be:
- now or never returns None
- message is sent 
- drop is called
- message is found and sent, being reordered

Restioson: Therefore, this is better in most but not all cases than just dropping

thomaseizinger: Sorry, I don't fully understand.

I think this test proves that, once you have a `ReceiveFuture`, it can complete in a single poll, even if it has never been polled yet. Isn't this enough?

I guess we are not testing the case in which it is `Waiting`.

thomaseizinger: What do you think of other test I just added?

Restioson: >  think this test proves that, once you have a ReceiveFuture, it can complete in a single poll, even if it has never been polled yet. Isn't this enough?

I agree, but this is not the race which I was describing. I might be able to write a test tomorrow showing exactly what I mean, though not with now_or_never, but with manual polling emulating it. 

I think the other test is also valuable, but doesn't quite test what I meant.

thomaseizinger: I think I understand what you are trying to showcase but I am not sure it is necessary.

As it stands, we know that messages can be re-ordered if the `ReceiveFuture` is dropped. That is also documented on the type and users can decide to accept this reordering and simply drop a `ReceiveFuture`.

If they really care about ordering, they can do what we do in `Context::select` and call `now_or_never` on the future.

The tests we add here prove that, no matter what state the `ReceiveFuture` is in (newly constructed or already polled), it will always yield the message IF it has one stored internally. With this test passing, we should be able to rely on `now_or_never` to basically be a 1-to-1 replacement for `cancel`.

Restioson: I think it comes down to what kind of ordering we would like to guarantee when multiple actors are involved. `now_or_never` has equivalent ordering with `Drop` *and even* `cancel` if there is only one receiving actor.
 
 ```rust
 let (addr, mut ctx) = Context::new(None);

let mut a = Cow { name: "Alfred" };

let mut rx_a = ctx.next_message();
let mut fut_ctx = std::task::Context::from_waker(noop_waker_ref());

// Pretend this is the poll in now_or_never...
assert!(rx_a.poll_unpin(&mut fut_ctx).is_pending());

// This will fulfill the waiting rx in rx_a
assert!(addr.send(Hello("first message")).split_receiver().now_or_never().is_some());

// This will go to the queue
assert!(addr.send(Hello("second message")).split_receiver().now_or_never().is_some());

// Pretend this receive is from another actor...
ctx.tick(ctx.next_message().await, &mut a).await;

// Pretend this is the drop in now_or_never
drop(rx_a);

// This receive is also from the other actor...
ctx.tick(ctx.next_message().await, &mut a).await;
```

In this case, the second actor has now handled the message out of order.

I don't actually think this is much of a problem, on second thought, but we should mention how ordering interacts with multiple actors, just in case, although I do not expect multiple actors on the same address to be relying on external message ordering.

thomaseizinger: I see. What is missing for me in the above case is to do something with `rx_a` after we have initially polled it. I think it is okay for things to behave weirdly if you have half-polled futures around that are not interacted with.

1. If you `Drop` the `rx_a` future before the first send, it doesn't have a message and thus no re-ordering takes place
2. If you `Drop` it afterwards, the message gets re-inserted and re-ordering might take place
3. If you poll it again after the first send (with `now_or_never`), the message will be delivered to you and you can pass it to an actor.

(3) is what we are doing in the `select` function now which I think is functionally equivalent to what we had with `cancel`.



> I don't actually think this is much of a problem, on second thought, but we should mention how ordering interacts with multiple actors, just in case, although I do not expect multiple actors on the same address to be relying on external message ordering.

I don't think we want to make any guarantees in regards to message ordering with our work-stealing scheduling. That would be kind of bizarre.

Restioson: > I think it is okay for things to behave weirdly if you have half-polled futures around that are not interacted with.

Fair enough. There was a similar discussion in flume at some point, if I remember correctly, and we came to a similar decision.

> If you poll it again after the first send (with now_or_never), the message will be delivered to you and you can pass it to an actor.

Technically, the message will be delivered to you whether or not you poll it more than once. As soon as you poll it once it will be delivered to you if the sender is given an opportunity.

> I don't think we want to make any guarantees in regards to message ordering with our work-stealing scheduling. That would be kind of bizarre.

Agreed.

thomaseizinger: Okay. I think this is ready to merge then, right?

Restioson: I think so, yes. Maybe as part of #121 we can mention the ordering

PR #140, opened 2022-07-17
Introduce optional `sink` feature

thomaseizinger:
Not every user may need the `sink` feature and we can minimise
our dependency tree in those cases.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,8 +16,8 @@ rust-version = "1.60.0"
 async-trait = "0.1.56"
 catty = "0.1.5"
 futures-core = { version = "0.3.21", default-features = false, features = ["alloc"] }
-futures-sink = { version = "0.3.21", default-features = false }
-futures-util = { version = "0.3.21", default-features = false, features = ["sink", "alloc"] }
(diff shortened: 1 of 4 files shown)


Restioson (commented):
I like this change! One thing I do wonder is whether it should be default or not?

thomaseizinger:
I am not a fan of default features. In my opinion, they are bad because too many library developers forget to disable default features they are not using which leads to bloated dependency trees.

The only exception is probably features like `std`. Speaking of which, we should be able to make `xtra` no-std friendly by just depending on `alloc` right? I've never written a no-std crate. Is being "no-std" by default a good thing?

Restioson:
> I am not a fan of default features. In my opinion, they are bad because too many library developers forget to disable default features they are not using which leads to bloated dependency trees.

Fair enough. We just need to make sure the fact that this functionality is moved is properly communicated. A note in the readme as here is probably ok

> Speaking of which, we should be able to make xtra no-std friendly by just depending on alloc right? 

Unfortunately not, since we depend on std's mutexes.

> Is being "no-std" by default a good thing?

I would say so! I'm not sure it's possible here, unfortunately.

PR #141, opened 2022-07-17
Inline `Sender` into `Address`

thomaseizinger:
Another step towards #126.

What is interesting about this patch-set is that it removes the reference count from `SendFuture` and `BroadcastFuture`. I think this is okay?

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -162,25 +162,25 @@ impl<A, Rc: RefCounter> Address<A, Rc> {
         message: M,
     ) -> SendFuture<
         <A as Handler<M>>::Return,
-        NameableSending<A, <A as Handler<M>>::Return, Rc>,
+        NameableSending<A, <A as Handler<M>>::Return>,
         ResolveToHandlerReturn,
     >
     where
         M: Send + 'static,
         A: Handler<M>,
     {
-        SendFuture::sending_named(message, self.0.clone())
+        SendFuture::sending_named(message, self.0.inner.clone())
     }
 
     /// Send a message to all actors on this address.
     ///
     /// For details, please see the documentation on [`BroadcastFuture`].
-    pub fn broadcast<M>(&self, msg: M) -> BroadcastFuture<A, M, Rc>
+    pub fn broadcast<M>(&self, msg: M) -> BroadcastFuture<A, M>
     where
         M: Clone + Sync + Send + 'static,
         A: Handler<M, Return = ()>,
     {
-        BroadcastFuture::new(msg, self.0.clone())
+        BroadcastFuture::new(msg, self.0.inner.clone())
     }
 
     /// Waits until this address becomes disconnected. Note that if this is called on a strong
--- a/src/address.rs
+++ b/src/address.rs
@@ -6,6 +6,7 @@ use std::fmt::{self, Debug, Formatter};
 use std::future::Future;
 use std::hash::{Hash, Hasher};
 use std::pin::Pin;
+use std::sync::Arc;
 use std::task::{Context, Poll};
 
 use event_listener::EventListener;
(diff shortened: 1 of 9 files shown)


src/broadcast_future.rs
```diff
@@ -47,11 +43,13 @@ where
     pub fn priority(self, priority: u32) -> Self {
         match self.inner {
             Inner::Initial {
-                message, sender, ..
+                message,
+                chan: sender,
```
thomaseizinger, line 47: Note to self: Remove this extra binding.

thomaseizinger:
> > What is interesting about this patch-set is that it removes the reference count from SendFuture and BroadcastFuture. I think this is okay?
> 
> I think that this would cause unintuitive behaviour. For instance, I'd expect the following to always send:
> 
> ```rust
> let addr: Address<MyActor, Strong> = addr; // just asserting this for test purposes
> tokio::task::spawn(addr.send(MyMessage));
> drop(addr);
> ```

Fair enough, so we need to keep a strong reference around in the `SendFuture`.

src/inbox/tx.rs
```diff
@@ -139,24 +26,24 @@ impl<A, Rc: TxRefCounter> SetPriority for SendFuture<A, Rc> {
 }
 
 #[must_use = "Futures do nothing unless polled"]
-pub enum SendFuture<A, Rc: TxRefCounter> {
+pub enum SendFuture<A> {
     New {
         msg: SentMessage<A>,
-        tx: Sender<A, Rc>,
+        chan: Arc<Chan<A>>,
```
Restioson, line 32: I guess we should just replace this with `tx` again?

thomaseizinger: You mean the variable name?

Restioson: I mean instead of holding the inner channel, we can just clone the sender, which is just chan + rc, right?

src/address.rs
```diff
@@ -269,23 +291,33 @@ impl<A, Rc: RefCounter> Eq for Address<A, Rc> {}
 /// address will compare as greater than a weak one.
 impl<A, Rc: RefCounter, Rc2: RefCounter> PartialOrd<Address<A, Rc2>> for Address<A, Rc> {
     fn partial_cmp(&self, other: &Address<A, Rc2>) -> Option<Ordering> {
-        Some(match self.0.inner_ptr().cmp(&other.0.inner_ptr()) {
-            Ordering::Equal => self.0.is_strong().cmp(&other.0.is_strong()),
-            ord => ord,
-        })
+        Some(
+            match Arc::as_ptr(&self.inner).cmp(&Arc::as_ptr(&other.inner)) {
```
Restioson, line 295: Just to check, does this properly compare data ptrs and not vtables?

Restioson (approved):
Thanks! Just those two things that I mentioned to address now

src/inbox/tx.rs
thomaseizinger: There is no more `Sender` with this PR :)

Restioson: Ah, right! It would be `Address` then, I guess!

thomaseizinger: I guess we could embed the `Address` yeah. Although that would be a cyclic dependency. It is not a problem but conceptually weird.

I am doing some other work on `SendFuture` atm so I'll go think about this once that is done.

src/address.rs
thomaseizinger: According to the docs, yes: https://doc.rust-lang.org/std/sync/struct.Arc.html#method.ptr_eq

thomaseizinger: Interesting, I guess this also needs to be changed then.

thomaseizinger:
I am going to close this for now, want to get some of the internal channel refactorings over the line first.

PR #142, opened 2022-07-17
Always store message priority in envelope

thomaseizinger:
Previously, we only stored the priority of broadcast messages in
the envelope. For regular messages, the priority was stored
outside in a wrapper struct.

This disparity is hard to understand as there is no apparent reason
why the priority could not be stored in the message envelope.

With this patch, we move the priority into the message envelope
and also expose a `set_priority` function for each envelope type
which allows us to consolidate the way we set priority from the
`SendFuture`.

### DIFF
--- a/src/inbox.rs
+++ b/src/inbox.rs
@@ -19,8 +19,7 @@ use crate::inbox::rx::{RxStrong, WaitingReceiver};
 use crate::inbox::tx::{TxStrong, WaitingSender};
 
 type Spinlock<T> = spin::Mutex<T>;
-pub type MessageToOneActor<A> = Box<dyn MessageEnvelope<Actor = A>>;
-type BroadcastQueue<A> = Spinlock<BinaryHeap<MessageToAllActors<A>>>;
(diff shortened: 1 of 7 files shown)


src/inbox.rs
```diff
@@ -196,7 +202,7 @@ enum MessageType {
 }
 
 pub enum SentMessage<A> {
-    ToOneActor(PriorityMessageToOne<A>),
```
thomaseizinger, line 199: I've removed the type alias because it results in a nice symmetry here.

thomaseizinger:
> Shouldn't we be able to store the priority as a normal `u32` and use `Arc::get_mut` and take mutable references for the boxed messages?

I will try! Didn't know about that method :)

thomaseizinger:
I've 

> Shouldn't we be able to store the priority as a normal `u32` and use `Arc::get_mut` and take mutable references for the boxed messages?

This works which is much nicer! I've also resolved the conflicts with master so this is ready to be reviewed again :)

PR #143, opened 2022-07-18
Rename message -> message_type and use Display for strs

Restioson:
Weird behaviour is exhibited by some subscribers with `message` as a field name, probably since it is reserved, so `message_type` is used instead. Also, the display implementation of type name strings are used to avoid unnecessary quotes in the output.

### DIFF
--- a/src/envelope.rs
+++ b/src/envelope.rs
@@ -88,20 +88,20 @@ impl Instrumentation {
     }
 
     #[cfg_attr(not(feature = "instrumentation"), allow(unused_variables))]
-    fn started<A>(message: &'static str) -> Self {
+    fn started<A>(message_type: &'static str) -> Self {
         #[cfg(feature = "instrumentation")]
         {
             let parent = Span(tracing::debug_span!(
                 "xtra_actor_request",
-                actor = std::any::type_name::<A>(),
-                %message,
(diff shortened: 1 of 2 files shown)


src/envelope.rs
```diff
@@ -88,20 +88,20 @@ impl Instrumentation {
...
     #[cfg_attr(not(feature = "instrumentation"), allow(unused_variables))]
-    fn started<A>(message: &'static str) -> Self {
+    fn started<A>(message_type: &'static str) -> Self {
         #[cfg(feature = "instrumentation")]
         {
             let parent = Span(tracing::debug_span!(
                 "xtra_actor_request",
-                actor = std::any::type_name::<A>(),
-                %message,
+                actor = %std::any::type_name::<A>(),
```
thomaseizinger, line 96: For consistency, should we rename this to `actor_type`?

thomaseizinger (approved):

thomaseizinger:
I've taken a stab at a better PR title and consequently commit message, feel free to change back if you don't agree.

Restioson:
Thanks!

src/envelope.rs
Restioson: Good point

PR #147, opened 2022-07-18
Unify message sending into one `SendFuture`

thomaseizinger:
This one future now handles:

- Sending a single message with the actor type being named
- Sending a single message with the actor type being erased
- Sending a broadcast message with the actor type being named

We can later use this to implement broadcasting for message channels but that requires a bit more work and I wanted to land this by itself.

Note that this also includes #105.


src/send_future.rs
```diff
@@ -26,250 +23,353 @@ use crate::{inbox, Error, Handler};
 /// In case an actor's mailbox is bounded, [`SendFuture`] will yield `Pending` until the message is queued successfully.
 /// This allows an actor to exercise backpressure on its users.
 #[must_use = "Futures do nothing unless polled"]
-pub struct SendFuture<R, F, TResolveMarker> {
-    inner: SendFutureInner<R, F>,
-    phantom: PhantomData<TResolveMarker>,
+pub struct SendFuture<F, TState> {
+    sending: F,
+    state: TState,
```
Restioson, line 28: Is the "T" representing "Type" still relevant here, since now these can carry data?

thomaseizinger: Probably, not. Happy to rename it just to `S` if you want!

Restioson (approved):
Thanks! This is really nice!

PR #148, opened 2022-07-18
Don't pass message name into envelope

thomaseizinger:
An envelope knows which message it is carrying, no need to pass
it in.


Restioson (approved):
This is a nice simplification, thanks!

PR #149, opened 2022-07-18
Encapsulate more functionality in `WaitingReceiver` 

thomaseizinger:
This is a meant to be a pure refactoring.

The primary motivation was to make the code easier to understand. With this patch-set, we introduce the concept of a `FulfillHandle` which is the counter part to `WaitingReceiver`. The actual state is now stored in `WaitingState`. This allows us to have richer APIs on the various types and we can use the type system to express that a `WaitingReceiver` and a `FulfillHandle` always come as a pair. Previously, we were just handling `Arc`s where it was not obvious that there is at most 1 strong and 1 weak count at any given time.

I took the opportunity to also add more documentation around what is happening here.


thomaseizinger:
Side note: This was a very insightful journey. I am beginning to more and more understand the details of the channel implementation here :)

src/inbox/waiting_receiver.rs
```diff
@@ -0,0 +1,169 @@
...
+use crate::envelope::MessageEnvelope;
+use crate::inbox::rx::Receiver;
+use crate::inbox::rx::RxRefCounter;
+use crate::inbox::ActorMessage;
+
+/// A [`WaitingReceiver`] is handed out by the channel any time [`Chan::try_recv`] is called on an empty mailbox.
+///
+/// [`WaitingReceiver`] implements [`Future`] which will resolve once a message lands in the mailbox.
+pub struct WaitingReceiver<A> {
+    state: Arc<spin::Mutex<WaitingState<A>>>,
```
Restioson, line 13: This is basically a oneshot, actually. I wonder whether this could lead to any greater simplification or not

thomaseizinger: I can try replacing it with catty if you want? We already depend on that.

thomaseizinger: However, I think it might be possible to extract our own little oneshot implementation where the sender is always a `Weak`.

thomaseizinger: I may attempt that once I've gotten around to do the same thing for `WaitingSender`. That one is a bit more complication to change so I didn't include it in here.

Restioson: > catty doesn't use a Weak internally but two strong references so it is not quite functionally equivalent.

I think it ends up being equivalent? Catty has the same semantics where if either side drops, it disconnects.

thomaseizinger: > > catty doesn't use a Weak internally but two strong references so it is not quite functionally equivalent.
> 
> I think it ends up being equivalent? Catty has the same semantics where if either side drops, it disconnects.

Good point. We can do that in a follow-up, right? The more important change here is to set the API and make things like `WakeReason` an implementation detail. How we realise the implementation is then free to change!

src/inbox/rx.rs
```diff
@@ -67,147 +75,98 @@ impl<A, Rc: RxRefCounter> Drop for Receiver<A, Rc> {
...
+                            waiting_receiver: waiting,
+                        });
+                    }
+                },
+                ReceiveFuture::Waiting(mut waiting) => match waiting.poll_unpin(cx) {
+                    Poll::Ready(Ok(msg)) => return Poll::Ready(msg),
+                    Poll::Ready(Err(rx)) => {
+                        // False positive wake up, try receive again.
+                        *this = ReceiveFuture::New(rx);
+                        continue;
```
thomaseizinger, line 151: I think we can spare this.

Restioson: What do you mean by spare?

src/inbox/waiting_receiver.rs
Restioson: I agree, yea

thomaseizinger:
I am not planning on adding anything else to this so feel free to merge this if you think it is ready :)

src/inbox.rs
```diff
@@ -81,41 +83,45 @@ impl<A> Chan<A> {
         let mut inner = self.chan.lock().unwrap();
 
         let result = match message {
-            SentMessage::ToAllActors(m) if !self.is_full(inner.broadcast_tail) => {
+            SentMessage::ToAllActors(m) => {
+                if self.is_full(inner.broadcast_tail) {
+                    // on_shutdown is only notified with inner locked, and it's locked here, so no race
```
Restioson, line 88: Comment is a bit odd here?

thomaseizinger: It is not my comment, it was already there, previously on line 89. I don't fully understand it to be honest so I just left it with the same code as it was before.

thomaseizinger: It appeared here: https://github.com/Restioson/xtra/pull/113/files#diff-5a063c294075054486f1cce09b584344bdea5f8ddec0f4551159f038318908d3R46

Restioson: Huh. I think it can be removed now. Probably a copy-paste error on my part

thomaseizinger, line 88:
```suggestion
```

PR #150, opened 2022-07-19
Move capacity from `Chan` to `ChanInner`

thomaseizinger:
This allows us to avoid some back and forth passing of data which
makes a few functions have less arguments and overall better
ergonomics.


Restioson (approved):
Happy for you to merge this after conflicts are resolved

PR #151, opened 2022-07-19
Use `catty` to implement `WaitingReceiver`

thomaseizinger:
This works well and saves about 60 LoC.


Restioson:
Are you sure we need this patch?

PR #153, opened 2022-07-20
Remove dependency on catty patch

Restioson:


thomaseizinger (approved):
I guess that works too.

I was just confused about the `&mut self` when you have interior mutability anyway!

Restioson:
> I was just confused about the &mut self when you have interior mutability anyway!

I think I did this since the future and receiver are the same thing, and I didn't want it to get polled by two different things at the same time and potentially mess up the waker.

PR #154, opened 2022-07-22
Use or_current when creating spans

Restioson:
This makes sure that if the xtra_actor_request is disabled, the other spans and child-spans of xtra_message_handler will be children of the would-be parent of xtra_actor_request, rather than being parentless, top-level spans.


thomaseizinger (commented):
Can we accommodate this with a test?

Restioson:
I've added a test and verified that it fails on master but passes here.

PR #155, opened 2022-07-22
Remove duplicated code for popping from broadcast channel

thomaseizinger:


src/inbox.rs
```diff
@@ -125,10 +125,11 @@ impl<A> Chan<A> {
         &self,
         broadcast_mailbox: &BroadcastQueue<A>,
     ) -> Result<ActorMessage<A>, WaitingReceiver<A>> {
-        let mut broadcast = broadcast_mailbox.lock();
-
-        // Peek priorities in order to figure out which channel should be taken from
-        let broadcast_priority = broadcast.peek().map(|it| it.priority());
+        // lock braodcast mailbox for as short as possible
```
Restioson, line 128: nit: typo

src/inbox.rs
```diff
@@ -148,13 +149,7 @@ impl<A> Chan<A> {
...
             }
             // Shared priority is less - take from broadcast
-            Ordering::Less => {
-                let msg = broadcast.pop().unwrap().0;
-                drop(broadcast);
-                inner.try_advance_broadcast_tail();
-
-                Ok(msg.into())
-            }
+            Ordering::Less => Ok(inner.pop_broadcast(broadcast_mailbox).unwrap().into()),
```
Restioson, line 152: I'm not sure we can do this, since the broadcast priority could be out of date, as the channel could get modified between when we check it and when we take from it

thomaseizinger: Ah, that makes sense. I can lock it at the top now though I think. That wasn't possible before.

thomaseizinger: Damn, doesn't work actually because we lock it again inside.

thomaseizinger: If we move the locking of `ChanInner` further up, I don't think there can be a race condition though. We only send into the broadcast mailbox with `ChanInner` locked.

Restioson: Yea, but wouldn't another receive future on the same mailbox be able to take out a message? Concurrent sends are not a problem since they can never decrease the max element of the heap.

thomaseizinger: I don't think so. The `ReceiveFuture` just calls `try_recv` which locks `ChanInner` first thing. How would a concurrent receive be possible?

Restioson: Sorry for getting back to this late - I think I was under the impression that we still only supplied the waiting rx with a notification to check its mailbox rather than the message itself

Issue #43, opened 2021-10-06
"message_cap" channel creation option should be more discoverable in documentation.


vi:
It is hard to know whether Xtra uses bounded or unbounded channels and how to set them up. Searching documentation for "bounded" does not lead to any hints.

It is hard to get to `xtra::Actor::create` when expecting this things to be in `xtra::address` or `xtra::message_channel`. No other mention about bounded vs unbounded seems to be in the documentation.

README example has `None` for creating actor, yet it is not clear that it means unbounded.

I suggest:

* Mention words "bounded" and "unbounded" explicitly in `xtra::Actor::create` doccomment. There is "unboudned", but not "bounded", so searching may fail to find it.
* Maybe add some rustdoc searching aliases, so that if user types "bounded" or "unbounded" in search box, he would see `create` method.
* Mention ability to use both channel types in READE, in the features list.
* Add a comment in README's example about `None` meaning not "default settings" (like users may expact), but "unbound mailbox".
* Maybe warn about blocking vs unblocking distinction in `xtra::Address::send`. `do_send` seems to already carry the warning, but based on `do_send`s doccomment alone, it may occur that xtra supports _only_ bound channels. It is hard to get from `do_send`'s "If the actor’s mailbox is full, it will block." to `Actor::create`'s `message_cap`.
* In `xtra::message_channel`, mention that the word "channel" is used specially in xtra and hint that user may want to search for the word "mailbox" instead.
* Maybe mention the word "channel" in `xtra::Actor::create`. Like like "... given the cap for the actor’s mailbox (i.e. channel)".

Restioson:
Good point, I will make sure to clarify this ahead of 0.6.0

thomaseizinger:
In case https://github.com/Restioson/xtra/pull/133 lands, we can rename the constructor and have `Mailbox::bounded` and `Mailbox::unbounded` which should certainly fix this.

Restioson:
Closing as #133 will land soon! :)

thomaseizinger:
> Closing as #133 will land soon! :)

I had it linked to the PR already :D

PR #156, opened 2022-07-24
Redesign xtra's spawn API

### LINKED ISSUE #136: Remove `ActorManager` in favor of executor specific modules / functions
Currently, we have `Actor::create` and the `ActorManager` struct which allows us to conveniently spawn actors in a single line onto a particular runtime:

```rust
let address = MyActor.create(None).spawn_global();
```

I am proposing to remove the `create` function and `ActorManager` in favor of executor-specific modules that will fulfill the same job. These APIs are of similar vein as #133.

```rust
let address = xtra::tokio::spawn_unbounded(MyActor);
let address = xtra::tokio::spawn_bounded(MyActor, 5);
let address = xtra::wasm_bindgen::spawn_bounded(MyActor, 5);
```

A slightly alternative API could be to make the user pass the `Mailbox` in which would reduce the cardinality a bit:

```rust
let address = xtra::tokio::spawn(MyActor, Mailbox::unbounded());
let address = xtra::tokio::spawn(MyActor, Mailbox::bounded(5));
let address = xtra::wasm_bindgen::spawn(MyActor, Mailbox::bounded(5));
```

Another variation could be to have more top-level functions instead of modules. Arguably, if there is only one function per module, it might not be worth it.

```rust
let address = xtra::spawn_tokio(MyActor, Mailbox::unbounded());
let address = xtra::spawn_tokio(MyActor, Mailbox::bounded(5));
let address = xtra::spawn_wasm_bindgen(MyActor, Mailbox::bounded(5));
```

Personally, I think I prefer the last one because it has a low cardinality, is expressive but also not too verbose.

thomaseizinger:
This removes `Actor::create`, `ActorManager` and the various runtime
extension traits in favor of top-level free functions for spawning
an actor instance into a specific runtime.

This is deemed to be more ergonomic because:

- Users do not need to import an extension trait for their runtime
- Users do not need to import the `Actor` trait to run it which was
otherwise required to call `create`.
- The implementation has less indirection, thus making it easier for
users to understand and perhaps adapt / write their own `spawn` fn
depending on their needs.

The implementation is not strictly functionally equivalent to what
we had before. In particular, we no longer support spawning in a
runtime handle of the tokio and smol runtimes. We accept this
regression because we assume that it is a niche usecase and the now
easier implementation should showcase users, how they can easily
write themselves a helper function that spawns an actor using a
runtime handle in a single line of code. In other words, the `spawn`
API of xtra is purposely design to make 80% of usecases work. For
everything else, users should write their own `spawn` function.

Closes #136.


src/spawn.rs
```diff
@@ -1,166 +1,63 @@
...
-pub use smol_impl::*;
+/// Spawns the given actor into the tokio runtime, returning an [`Address`] to it.
+///
+/// The 2nd parameter is the mailbox size with `None` creating an unbounded mailbox.
 #[cfg(feature = "tokio")]
-pub use tokio_impl::*;
-#[cfg(feature = "wasm_bindgen")]
-pub use wasm_bindgen_impl::*;
+#[cfg_attr(docsrs, doc(cfg(feature = "tokio")))]
+pub fn spawn_tokio<A>(actor: A, mailbox_size: Option<usize>) -> crate::Address<A>
```
thomaseizinger, line 6: With #133, I'd like to make this parameter the actual mailbox and not just mailbox size so we can also fix #43 by making the constructors on `Mailbox` say `bounded` and `unbounded`.

For now, I chose to just take the size here because I figured it is an easier transition overall.

src/spawn.rs
```diff
@@ -1,166 +1,63 @@
...
 
+/// Spawns the given actor into the async_std runtime, returning an [`Address`] to it.
+///
+/// The 2nd parameter is the mailbox size with `None` creating an unbounded mailbox.
 #[cfg(feature = "async_std")]
-mod async_std_impl {
-    use super::*;
-    use crate::{Actor, ActorManager, Address};
+#[cfg_attr(docsrs, doc(cfg(feature = "async_std")))]
+pub fn spawn_async_std<A>(actor: A, mailbox_size: Option<usize>) -> crate::Address<A>
```
thomaseizinger, line 22: I theory, these functions could be generated by a macro but it is only a handful of functions so I didn't want to bother :)

src/spawn.rs
Restioson: That makes sense!

src/spawn.rs
```diff
@@ -1,166 +1,63 @@
...
+where
+    A: crate::Actor<Stop = ()>,
+{
+    let (address, context) = crate::Context::new(mailbox_size);
+
+    tokio::spawn(context.run(actor));
 
-/// An `Spawner` represents anything that can spawn a future to be run in the background. This is
-/// used to spawn actors.
-pub trait Spawner {
```
Restioson, line 14: How would you feel about an alternate API like
```rust
fn spawn<A: Actor, R: Spawner>(act: A, rt: R) -> Address<A> {
    
    let (address, context) = crate::Context::new(mailbox_size);
    rt.spawn(context.run(actor));
    address
}
```
`Spawner` could be implemented for `tokio::runtime::Handle`, and also for a variety of unit structs such as `TokioGlobal`, `SmolGlobal`, `WasmLocal`, etc, and it could be called like `xtra::spawn(act, xtra::rt::TokioGlobal)`. The advantage here would be that we could cut down on the number of functions in the public API, and it might simplify things from an understanding perspective - just swap out the `rt` arg to change where it's spawned to

thomaseizinger: I am not sure the indirection is worth it? Typically, a project will only use one kind of runtime and having to specify the runtime as an argument is always another type you need to type / import.

In reality, this argument is always going to be a constant because of the same reason so it feels like expressing it via functions is better than arguments (which are typically only used for things that can be dynamic).

I've also mentioned it in the PR description: These functions are an 80/20 solution to me: Serve 80% of usecases with 20% of the effort. The implementation is so simple that a project with different needs can write one in 2 minutes. In fact, I'd expect most projects to eventually have their own `spawn_actor` function that standardizes the mailbox size and runtime for the project.

thomaseizinger: If it wouldn't be such an ergonomic hit, I'd actually love to remove any runtime integration so we are truly runtime agnostic. Unfortunately, I don't think that is possible if we want to provide a way of spawning an actor in a single line of code.

Restioson: > If it wouldn't be such an ergonomic hit, I'd actually love to remove any runtime integration so we are truly runtime agnostic. Unfortunately, I don't think that is possible if we want to provide a way of spawning an actor in a single line of code.

Yea, AFAIK there is currently no way to abstract over all runtimes unless you do it yourself.

thomaseizinger: > The advantage here would be that we could cut down on the number of functions in the public API, and it might simplify things from an understanding perspective - just swap out the `rt` arg to change where it's spawned to

Whilst we cut down the number of functions in the public API, the number of public API elements is actually larger:

- `Spawner`
- `spawn`
- 1 struct per runtime, i.e. 4

That is 6 elements vs 4 in the proposed solution, plus, `spawn` has 1 more parameter and it has here.

I am not sure if number of API elements is a good metric here but I think the increased number of moving parts makes it a bit harder to understand.

thomaseizinger: We could provide both, release an RC and ask users, which one they prefer?

Restioson: That makes sense, I think. Happy to merge this for now though and we can maybe add the other API in a followup?

thomaseizinger: Cool! It is ready from my end :)

PR #158, opened 2022-07-28
Introduce generic `ChanPtr`

thomaseizinger:
This introduces a generic `ChanPtr` type which adds reference counting to the channel.

This allows us to remove the `Sender` type and promote `inbox::Receiver` to a (future public) `Mailbox` type.

On most parts, the implementation is the same. Important differences are:

- The actual logic for reference counting is encapsulated on `Chan`. This used to be #162 but I ended up doing it here because the merge conflicts would have been insane.
- ~`RefCounterPolicy` does not have an `into_either` fn. This is solved by special-casing `impl` blocks for `ChanPtr`.~
- The new implementation is accompanied by a range of tests.
- ~We remove the `RefCounter` trait from the public `ref_count` module. This is not actually necessary and allows us to avoid the sealing of the trait because users cannot name it anyway.~


thomaseizinger:
If we end up merging https://github.com/Restioson/xtra/pull/160, I am gonna iterate on this design again. I think it would be a step backwards to introduce a "ref count policy" again just so we can use this abstraction also for the receiver side.

thomaseizinger:
One alternative idea to this would be to retain the `Sender` and `Receiver` newtypes but implement `Deref` on them so we don't need to forward all the methods.

Restioson:
I think this will need to be rebased now that #160 has been merged

thomaseizinger:
> I think this will need to be rebased or updated now that #160 has been merged

Yep, I'll need to think about a new design here.

thomaseizinger:
So after thinking about this more, I think it is overall better to actually re-introduce a ref-count policy with this PR. Here is my thinking:

I am currently trying to achieve symmetry between various modules. #159 will introduce a top-level `recv_future` module that is equivalent to the `send_future` module. Both of these futures need to have a reference-counted pointer to the inner channel, i.e. they can't just take `Arc<ChanA>>`.

`Address` wraps a `Sender` which is technically not necessary, all of the logic of `Sender` can go into `Address` _except_ that we would be introducing circular references between the `address` and the `send_future` module because `SendFuture` needs to have reference-counted pointer to `Chan` which - after the removal of `Sender` would be `Address`.

The same situation applies to `Receiver`. With #133, `Receiver` will be renamed to `Mailbox` or at least there will be a public `Mailbox` type with a `receive` function on it that will return a `ReceiveFuture`. `Mailbox` will thus be the equivalent of `Address`. `ReceiveFuture` will have to also contain a reference-counted pointer to the `Chan` which again, shouldn't be `Mailbox` itself because of the circular-reference issue.

Thus, I think despite an `Rx` reference count type not being strictly necessary, the `ChanPtr` abstraction is quite clean and provides a safe way for us to manage the reference counts to the channel. It can easily embedded in other types, most of the code can be reused between `Address` and `Mailbox` and the symmetry makes it easy to understand.

Restioson:
> except that we would be introducing circular references between the address and the send_future module 

Is this a problem? I don't think this will be a compile fail, anyway.

thomaseizinger:
> > except that we would be introducing circular references between the address and the send_future module
> 
> Is this a problem? I don't think this will be a compile fail, anyway.

It is not a technical problem but it is a design smell. With the right abstractions, you should be able to avoid circular dependencies and code is much easier understood if dependencies are uni-directional.

> Is this essentially what you are imagining?
> 
> ```rust
> type Address<A, Rc> = ChanPtr<A, Rc>; // maybe with newtypes
> type Mailbox<A> = ChanPtr<A, RxStrong>;
> ```
> 
> Then, the difference between RxStrong and TxStrong is simply which counter is modified.

This but with new-types so we have control over the public API of the library.

thomaseizinger:
Okay, this is ready for another review @Restioson! I've integrated #162 into here because the merge conflicts would have been insane.

I also updated the PR description to be an up-to-date summary of what this patch set does.

tests/basic.rs
```diff
@@ -873,14 +875,14 @@ fn message_channel_debug() {
     assert_eq!(
         format!("{:?}", mc),
         "MessageChannel<basic::Hello, alloc::string::String>(\
-            Sender<basic::Greeter> { rx_count: 1, tx_count: 1, rc: TxStrong(()) }\
+            ChanPtr<basic::Greeter, xtra::inbox::chan_ptr::TxStrong> { rx_count: 1, tx_count: 1 }\
```
Restioson, line 878: I wonder if we should be leaking this impl detail. This is just a general comment, though, as we were leaking Sender before anyway :smile:

src/address.rs
```diff
@@ -145,11 +158,6 @@ impl<A, Rc: RefCounter> Address<A, Rc> {
         self.len() == 0
     }
 
-    /// Convert this address into a generic address which can be weak or strong.
-    pub fn as_either(&self) -> Address<A, Either> {
```
Restioson, line 149: I'm not sure it's an advantage to take this off of Address<A, Either>, as it means that it might be harder to write code generic over all refcounters

src/lib.rs
```diff
@@ -36,10 +37,9 @@ pub mod prelude {
 /// This module contains types representing the strength of an address's reference counting, which
 /// influences whether the address will keep the actor alive for as long as it lives.
 pub mod refcount {
-    pub use crate::inbox::tx::TxEither as Either;
-    pub use crate::inbox::tx::TxRefCounter as RefCounter;
```
Restioson, line 40: I would like to keep this here, as it allows users to write generic code like
```rust
trait AddressExt { .. }
impl<A: Actor, Rc: RefCounter> AddressExt for Address<A, Rc> { .. }
```

Otherwise, you need to provide this only for `Address<A, TxWeak>`, which compels the use of `.into_either`

tests/basic.rs
thomaseizinger: No preference here really. Can we open an issue, add it to the milestone and defer that decision? :)

src/address.rs
thomaseizinger: The problem is that if we want to have it here, we need a conversion function on `RefCounter` again so I can call it from any ref counter.

Doable but I figured it is unnecessary. There are only two types of ref counter this is relevant for so it is effectively the same public API.

src/lib.rs
thomaseizinger: A wild requirement appears! 😁

I hadn't thought of this but it makes sense.

It would still be possible to abstract over all addresses, we'd just need to implement it multiple times but it is a bit repetitive.

Let me add a test/example for this in another PR so it's documented and I'll think about how to solve it here.

The unfortunate consequence of exporting the trait is we need to seal it, which adds complexity again :/

src/address.rs
thomaseizinger: This is of course different if we want users to be able to call this from an extension trait.

tests/basic.rs
Restioson: Sure!

src/lib.rs
Restioson: Hm... I wouldn't mind sealing it too much, to be honest, given that it's not a *super* huge extra complexity to handle. It is a little annoying, though.

src/address.rs
Restioson: I think regardless there should be a way, in the case of generic impls over RefCounter

src/lib.rs
thomaseizinger: > Let me add a test/example for this in another PR so it's documented and I'll think about how to solve it here.

https://github.com/Restioson/xtra/pull/169I've opened a PR here.

src/address.rs
thomaseizinger: I've reintroduced `RefCounter::into_either` in [`40fd855` (#158)](https://github.com/Restioson/xtra/pull/158/commits/40fd85596eafd4dc0b11dda55eac0d29ba60258a).

I've also opened https://github.com/Restioson/xtra/pull/171 which adds a test that these APIs should exist.

thomaseizinger:
All dependent PRs have been merged and feedback has been addressed :)

thomaseizinger:
Would appreciate another review here @Restioson :)

src/inbox/chan_ptr.rs
```diff
@@ -0,0 +1,402 @@
...
+impl RefCounter for TxEither {}
+impl RefCounter for Rx {}
+
+/// Private module to ensure [`RefCounter`] is sealed and can neither be implemented on other types
+/// nor can a user outside of this crate call functions from [`RefCounter`].
+mod private {
+    use super::*;
+
+    pub trait RefCounter {
+        type Either;
```
Restioson, line 214: Can this not directly refer to the `Either` type instead of being through an assoc type?

thomaseizinger: It can although it isn't super clean because `RefCounter` is more generic than `TxEither`. We can name the function `into_tx_either` and panic similar as we do now to avoid the associated type.

src/inbox/chan_ptr.rs
```diff
@@ -0,0 +1,402 @@
...
+            }
+        }
+
+        fn into_either(self) -> Self::Either {
+            self
+        }
+    }
+
+    impl RefCounter for Rx {
+        type Either = ();
```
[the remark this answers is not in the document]
thomaseizinger: Would you then only require `IntoEither` separately on `Address`? I can play around with that idea!

src/mailbox.rs
```diff
@@ -89,21 +66,25 @@ pub enum ReceiveFuture<A> {
...
 pub struct Waiting<A> {
-    channel_receiver: Receiver<A>,
+    channel: ChanPtr<A, Rx>,
+    broadcast_mailbox: Arc<BroadcastQueue<A>>,
     waiting_receiver: WaitingReceiver<A>,
 }
 
 impl<A> Future for Waiting<A> {
-    type Output = Result<ActorMessage<A>, Receiver<A>>;
+    type Output = Result<ActorMessage<A>, (ChanPtr<A, Rx>, Arc<BroadcastQueue<A>>)>;
```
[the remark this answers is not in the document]
thomaseizinger: The difference is in not having a circular dependency between `Waiting` and `Mailbox`.

thomaseizinger:
> I've left some comments - apologies for the long wait, I've been busy as term is coming to a close so we are writing tests and handing in assignments currently. It's vac soon so I should have more time then though :)

No worries and thanks for taking the time :)

src/inbox/chan_ptr.rs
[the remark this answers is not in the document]
Restioson: I think that would make sense yea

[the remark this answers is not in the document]
thomaseizinger: It actually turned out quite clean. Let me know what you think!

src/inbox/chan_ptr.rs
```diff
@@ -0,0 +1,385 @@
...
+
+/// A reference counter for the receiving end of the channel or in actor terminology, the mailbox
+/// of an actor.
+pub struct Rx(());
+
+/// Defines a reference counting policy for a channel pointer.
+///
+/// # Implementation note
+///
+/// This trait only exists because we cannot specialise `Drop` impls in Rust, otherwise, we could
```
Restioson, line 192: We should note the existence of Into<Either> here, as well as that it can be used to generalise over different RC types in impls

src/address.rs
```diff
@@ -83,15 +83,26 @@ impl<A> Address<A, Strong> {
...
     pub fn downgrade(&self) -> WeakAddress<A> {
-        Address(self.0.downgrade())
+        Address(self.0.to_tx_weak())
+    }
+}
+
+/// Functions which apply to any kind of address, be they strong or weak.
+impl<A, Rc> Address<A, Rc>
+where
+    Rc: RefCounter + Into<Either>,
```
Restioson, line 101: Not sure if it's worth encapsulating this bound into a new RC trait that we actually export or not - it provides the same experience as before but also there isn't much benefit outside of that. Maybe a note on the doc of RefCounter would be enough

thomaseizinger: I don't think it is worth it. Unless users call actually want to call `as_either`, the `Into` bound is not actually mandatory for the others. Perhaps we should demonstrate that in the public API test.

src/inbox/chan_ptr.rs
thomaseizinger: Can you make a suggestion via GitHub on what exactly you'd want to see? :)

src/inbox/chan_ptr.rs
```diff
@@ -0,0 +1,385 @@
...
+/// write:
+///
+/// - `impl Drop for ChanPtr<A, TxStrong> { }`
+/// - `impl Drop for ChanPtr<A, TxEither> { }`
+/// - `impl Drop for ChanPtr<A, Rx> { }`
+///
+/// and call the appropriate functions on `Chan`.
```
Restioson, line 199:
```suggestion
/// and call the appropriate functions on `Chan`.
///
/// It can also be used to generalise over the reference counter type in extension trait implementations on `Address`. For example:
/// ```
/// # use xtra::{Actor, Address, RefCounter};
/// trait MyExtension {
///    fn say_hi(&self) {
///        println!("Hi!"); // Presumably, you would do something useful here
///    }
/// }
///
/// impl<A, Rc> MyExtension for Address<A, Rc>
///    where A: Actor,
///         Rc: RefCounter
/// {}
```

Restioson: I am not sure if this is formatted correctly or compiles :sweat_smile:

src/address.rs
thomaseizinger: Added a new test to the public API module.

src/inbox/chan_ptr.rs
thomaseizinger: Cool, thanks! I'll fix it up :)

Restioson:
Good to merge once tests pass I think

thomaseizinger:
> Good to merge once tests pass I think

Needs your approval though :)

Restioson (approved):

PR #159, opened 2022-07-29
Introduce top-level `recv_future` module

thomaseizinger:
This creates consistency with the `send_future` module. To achieve this, we temporarily make the fields of `Receiver` public. This will go away once we merge #158 because it allows us to directly embed the `ChanPtr` and `BroadcastQueue` type in the `ReceiveFuture` instead of the actual `Receiver`.


thomaseizinger:
This PR actually made me wonder, why do we have the `RxWeak` type? We don't ever seem to use it :)

thomaseizinger:
> This PR actually made me wonder, why do we have the `RxWeak` type? We don't ever seem to use it :)

I had a crack at removing this in https://github.com/Restioson/xtra/pull/160.

thomaseizinger:
Setting this to draft until #160 is either merged or closed.

thomaseizinger:
Going to continue here once #158 is merged.

thomaseizinger:
I've merged https://github.com/Restioson/xtra/pull/158 into here under the assumption that it is not going to change much anymore until it is merged.

thomaseizinger:
The commit log is a bit of a message because of the squash merging that I did but the final diff is correct :D

Restioson (approved):

Restioson:
This is a pure refactor, right?

thomaseizinger:
> This is a pure refactor, right?

Yes this is a pure refactor. The primary change is that we have only one `ReceiveFuture` now which is part of the public API. Once we land https://github.com/Restioson/xtra/pull/133, `Mailbox` will be a public type and thus I thought it is better if we treat `ReceiveFuture` the same way we treat `SendFuture`.

Restioson:
Cool :) In that case, feel free to merge when ready

PR #160, opened 2022-07-29
Remove generic ref-counting from `Receiver`

thomaseizinger:
We never use weak references to mailboxes and don't expose a way for the user to construct one. Hence, the concept of generic reference counting is not useful here and only adds complexity.


Restioson (approved):
This is nice! At some point I think RxWeak was in use but clearly not anymore :smile:

PR #162, opened 2022-08-02
Encapsulate reference counting logic on `Chan`

thomaseizinger:
This follows the 'Tell, don't ask'-principle. Instead of first
decrementing the count on `Chan` and acting on it being a certain
value, we _delegate_ the decrementing to `Chan` and have it perform
the correct logic when necessary. This simplifies calling sites.

### DIFF
--- a/src/inbox.rs
+++ b/src/inbox.rs
@@ -55,11 +55,13 @@ impl<A> Chan<A> {
         }
     }
 
-    pub fn increment_receiver_count(&self) {
+    /// Callback to be invoked every time a receiver is created.
+    pub fn on_receiver_created(&self) {
         self.receiver_count.fetch_add(1, atomic::Ordering::Relaxed);
     }
 
-    pub fn decrement_receiver_count(&self) {
+    /// Callback to be invoked every time a receiver is destroyed.
+    pub fn on_receiver_dropped(&self) {
         // Memory orderings copied from Arc::drop
         if self.receiver_count.fetch_sub(1, atomic::Ordering::Release) != 1 {
             return;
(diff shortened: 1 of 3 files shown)


src/inbox.rs
```diff
@@ -70,6 +70,22 @@ impl<A> Chan<A> {
         self.shutdown_waiting_senders();
     }
 
+    pub fn increment_sender_count(&self) {
+        // Memory orderings copied from Arc::clone
+        self.sender_count.fetch_add(1, atomic::Ordering::Relaxed);
+    }
+
+    pub fn decrement_sender_count(&self) {
```
Restioson, line 78: I like this change overall, but this name I find to be a little misleading now - it has extra behaviour in some cases beyond just decrementing the refcount

src/inbox.rs
```diff
@@ -70,6 +70,22 @@ impl<A> Chan<A> {
         self.shutdown_waiting_senders();
     }
 
+    pub fn increment_sender_count(&self) {
```
Restioson, line 73: I think I originally did this directly on TxStrong so that it wouldn't seem like any reference to the Chan is able to increment the sender count (notably, weaks cant). Though, that wasn't exactly a very robust way of encapsulating that :sweat_smile:

src/inbox.rs
thomaseizinger: I am open to suggestions :)

We could name it after an event handler / hook, like `on_sender_dropped` and the other one `on_sender_created`?

src/inbox.rs
thomaseizinger: I get the idea!

I'd frame `Chan` as (powerful) low-level building blocks for the high-level concepts of `Address` and `Mailbox`. Thus, I'd say we favor encapsulation over a misuse-resistant interface. The latter is nice but because it is an interface private to the crate, I think it is acceptable.

Plus, tying the decrementing and shutting down of receivers together makes it much easier to understand the internals of `xtra`.

src/inbox.rs
Restioson: I think that makes sense!

src/inbox.rs
Restioson: Fair enough!

thomaseizinger:
Closing this because it is being integrated into #158.

PR #163, opened 2022-08-06
Add an `Actor` custom derive macro

### LINKED ISSUE #127: Add a custom-derive for `Actor`
Although it doesn't require much boilerplate at the moment, I think it would be a nice addition to add a `derive(Actor)` macro that generates a default `Actor` implementation with `Stop = ()`.

thomaseizinger:
Fixes #127.

### DIFF
--- a/macros-test/Cargo.toml
+++ b/macros-test/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "macros-test"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+xtra = { path = "../", features = ["macros"] }
+
+[dev-dependencies]
+trybuild = "1"
--- a/macros-test/Cargo.toml
+++ b/macros-test/Cargo.toml
@@ -4,8 +4,6 @@ version = "0.1.0"
 edition = "2021"
 publish = false
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
 [dependencies]
 xtra = { path = "../", features = ["macros"] }
 
--- a/macros-test/Cargo.toml
+++ b/macros-test/Cargo.toml
@@ -10,4 +10,4 @@ publish = false
 xtra = { path = "../", features = ["macros"] }
 
 [dev-dependencies]
-trybuild = "1"
+trybuild = "1.0.64"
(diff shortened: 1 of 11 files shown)


macros-test/tests/pass/basic.rs
```diff
@@ -0,0 +1,8 @@
+#[derive(xtra::Actor)]
+struct MyActor {
+
+}
```
thomaseizinger, line 4:
```suggestion
struct MyActor;
```

macros-test/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "macros-test"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
Restioson, line 7: Small nit :)

macros/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "xtra-macros"
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+proc-macro = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
Restioson, line 9: same vein as first nit

Restioson (approved):
I like this change in concept, but I do wonder if this is the most optimal directory structure/layout. I see that https://github.com/tokio-rs/tokio has a separate directory for the main tokio crate, whereas this has the crate root as `.`. This also doesn't use a workspace, I think.

macros-test/Cargo.toml
thomaseizinger: You'd want this comment removed or what is the nit? :D

thomaseizinger:
> I like this change in concept, but I do wonder if this is the most optimal directory structure/layout. I see that https://github.com/tokio-rs/tokio has a separate directory for the main tokio crate, whereas this has the crate root as `.`. This also doesn't use a workspace, I think.

We do use a workspace here, it is defined in the main crate!

I don't care much about the structure but if we want to change it, I'd suggest a different PR because it will be quite the diff noise :)

macros-test/Cargo.toml
Restioson: Yea I think it's best to leave it out as it's just fluff

thomaseizinger, line 7:
```suggestion
```

macros/Cargo.toml
thomaseizinger, line 9:
```suggestion
```

Restioson:
> We do use a workspace here, it is defined in the main crate!

Oh, I hadn't seen that!

PR #164, opened 2022-08-07
Allocate queues in channel according to capacity

thomaseizinger:
In case the user passes a capacity, we know how many elements we
will be storing in those queues. We can initialise them directly
with this capacity which avoids growing those buffers on demand.

### DIFF
--- a/src/inbox.rs
+++ b/src/inbox.rs
@@ -301,13 +301,15 @@ struct ChanInner<A> {
 
 impl<A> ChanInner<A> {
     fn new(capacity: Option<usize>) -> Self {
+        let pre_alloc_capacity = capacity.unwrap_or(0);
+
         Self {
             capacity,
-            ordered_queue: VecDeque::default(),
-            waiting_senders: VecDeque::default(),
-            waiting_receivers_handles: VecDeque::default(),
-            priority_queue: BinaryHeap::default(),
-            broadcast_queues: Vec::default(),
+            ordered_queue: VecDeque::with_capacity(pre_alloc_capacity),
+            waiting_senders: VecDeque::with_capacity(pre_alloc_capacity),
+            waiting_receivers_handles: VecDeque::with_capacity(pre_alloc_capacity),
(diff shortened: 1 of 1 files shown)


Restioson:
Hm, I would actually tend to disagree with this change. This adds a fixed cost of `8 * capacity` bytes per each type of mailbox, even if you don't use them. I intentionally left them as Vec::new as that doesn't allocate until use.

thomaseizinger:
> Hm, I would actually tend to disagree with this change. This adds a fixed cost of `8 * capacity` bytes per each type of mailbox, even if you don't use them. I intentionally left them as Vec::new as that doesn't allocate until use. 

On the other hand, with this change we'd have a much more predictable memory footprint as it wouldn't change during operation.

I am not sure what is better?

thomaseizinger:
If a user really operates on constrained hardware or wants to preserve memory, they could set capacity to 0 right?

thomaseizinger:
> If a user really operates on constrained hardware or wants to preserve memory, they could set capacity to 0 right?

I want to add a test that this actually works! Seems like an important enough case to cover.

Restioson:
If they set capacity to zero that'd affect the default priority queue too, though

thomaseizinger:
> If they set capacity to zero that'd affect the default priority queue too, though 

And? It effectively means that `split_receiver` is useless because you can't queue the message so `SendFuture` will yield until the actor is ready to process the message.

But other than that, `send` will block until the message is processed, same as before.

Am I missing something?

Restioson:
Ideally, the user shouldn't have to choose between getting to use `split_receiver` and not allocating extra capacity for a channel that is never used. In the spirit of Rust's zero-cost abstractions and pay-for-what-you-use, I'd like the extra channels to add as little overhead as possible if you don't use them.

Restioson:
It's also worth mentioning that the ordering behaviour might not work as expected with a rendezvous channel, but I'm not 100% sure of this

Issue #94, opened 2022-06-21
Re-consider separate, ordered mailbox


thomaseizinger:
With https://github.com/Restioson/xtra/pull/85, we are introducing 3 mailboxes: ordered, prioritized and broadcast.

Ordered exists because `BinaryHeaps` don't preserve insertion order and thus, messages with the same priority (like 0 as the default) could be processed in a different order, assuming they are used with `split_receiver` and the processing thus happens async.

I am claiming that this is way too much of a detail to expose to the user (which has to eventually learn that there are 3 mailboxes) so my suggestion would be to:

a) Use a data structure that preserves insertion order
b) Accept the re-ordering of messages

I am not data structure expert but couldn't we for example also use an atomic counter to tag each incoming message and sort the heap based on priority and message counter? That should preserve the sending order even among messages with the same priority.

Restioson:
> Couldn't we for example also use an atomic counter to tag each incoming message and sort the heap based on priority and message counter?

In theory, yes, but how do you propose to handle overflows? This was the main issue I ran into thinking about this option

thomaseizinger:
> > Couldn't we for example also use an atomic counter to tag each incoming message and sort the heap based on priority and message counter?
> 
> In theory, yes, but how do you propose to handle overflows? This was the main issue I ran into thinking about this option

Reset it every time the queue is empty perhaps?

Am I too naive in saying that we could also just not handle them? The "problem" would be a slight out of order handling of same-priority messages.

From experience, actors receive messages from various addresses concurrently, so the exact order in which they will be handled is unclear anyway so this wouldn't even be noticeable.

But if we accept this behaviour in an edge case, we might also just accept it all the time :D

thomaseizinger:
> b) Accept the re-ordering of messages

Expanding more on this:

Usage of `split_receiver` makes message delivery asynchronous and I think by definition, timing of events doesn't matter in an async system. (I've found [this](https://en.m.wikipedia.org/wiki/Asynchronous_system) but it doesn't have many follow-up references.)

In other words, I would argue that someone using `split_receiver` explicitly gives up control over the ordering of their messages.

We may want to debate the name of that function again because it doesn't communicate that right away, only conceptually.

thomaseizinger:
If we would implement https://github.com/Restioson/xtra/issues/92#issuecomment-1162267933, then this issue would go away because the `BinaryHeap` could be replaced by a `VecDeque` per message priority.

thomaseizinger:
[...]

PR #166, opened 2022-08-08
Differentiate single and broadcast message on type-system level

thomaseizinger:
Prior to this patch-set, we had constructs like `MessageType` and `SentMessage`. These were used to write code that is generic over a particular message type. In reality however, we don't actually need to differentiate between these cases because the codepaths for each type are statically known in all cases.

Unfortunately, introducing an actual split introduced a bit more code (net ~ 70 lines) but in exchange, we remove a few `unreachable` error clauses and make it overall easier to follow the dataflow through xtra's channel implementation.


src/send_future.rs
```diff
@@ -70,8 +70,11 @@ where
     }
 }
 
-/// "Sending" state of [`SendFuture`] for cases where the actor type is named.
-pub struct ActorNamedSending<A, Rc: RefCounter>(Sending<A, Rc>);
+/// "Sending" state of [`SendFuture`] for cases where the actor type is named and we sent a single message.
+pub struct ActorNamedSending<A, Rc: RefCounter>(Sending<A, MessageToOne<A>, Rc>);
+
+/// "Sending" state of [`SendFuture`] for cases where the actor type is named and we broadcast a message.
+pub struct ActorNamedBroadcasting<A, Rc: RefCounter>(Sending<A, MessageToAll<A>, Rc>);
```
thomaseizinger, line 77: I am not too happy with the terminology that we use to differentiate between regular messages and broadcasts. We a bit of a mix here throughout the codebase.

That said, I don't think this PR makes it any worse other than maybe surfacing this problem even more.

src/inbox.rs
```diff
@@ -493,7 +493,7 @@ impl<A> From<Box<dyn MessageEnvelope<Actor = A>>> for SentMessage<A> {
 }
 
 /// An error returned in case the mailbox of an actor is full.
-pub struct MailboxFull<A>(pub Arc<Spinlock<WaitingSender<A>>>);
+pub struct MailboxFull<A>(pub Arc<Spinlock<WaitingSender<SentMessage<A>>>>);
```
Restioson: Not a blocker for this merge or the release of 0.6, but it'd be really nice to be able to replace this with a rendezvous send

src/inbox.rs
```diff
@@ -566,7 +563,10 @@ impl<A> WaitingSender<A> {
     }
 }
 
-impl<A> Future for WaitingSender<A> {
+impl<M> Future for WaitingSender<M>
+where
+    M: Unpin,
```
Restioson: Is this bound going to limit what messages users can send?

src/inbox.rs
```diff
@@ -301,7 +312,10 @@ impl<A> Chan<A> {
 struct ChanInner<A> {
     capacity: Option<usize>,
     ordered_queue: VecDeque<Box<dyn MessageEnvelope<Actor = A>>>,
-    waiting_senders: VecDeque<Weak<Spinlock<WaitingSender<SentMessage<A>>>>>,
+    waiting_send_to_one:
+        VecDeque<Weak<Spinlock<WaitingSender<Box<dyn MessageEnvelope<Actor = A>>>>>>,
+    waiting_send_to_all:
+        VecDeque<Weak<Spinlock<WaitingSender<Arc<dyn BroadcastEnvelope<Actor = A>>>>>>,
     waiting_receivers_handles: VecDeque<FulfillHandle<A>>,
```
Restioson: It might be interesting (now or in another PR) to introduce 3 here - 1 per queue - which use the queue's type. So, the ordered queue gets a VecDeque, and priority queues get BinaryHeaps. This should preserve ordering behaviour for rendezvous channels.

src/inbox.rs
```diff
@@ -485,6 +466,35 @@ impl<A> ChanInner<A> {
     }
 }
 
+fn find_remove_highest_priority<M>(
```
Restioson, line 457: Related to: how I mentioned using BinaryHeap for priority messages, this could be removed & replaced with an O(log n) binary heap remove call

src/inbox.rs
```diff
@@ -293,6 +231,80 @@ impl<A> Chan<A> {
     }
 }
 
+pub trait TrySend<M> {
+    fn try_send(&self, message: M) -> Result<Result<(), MailboxFull<M>>, Error>;
```
Restioson, line 237: Isn't this essentially equivalent to what we had before, but based on a trait rather than an enum?

src/inbox.rs
thomaseizinger: Making this generic is a step towords that, is it not?

src/inbox.rs
thomaseizinger: It is functionally equivalent but a trait with type parameters is a compile-time concept and an enum is a runtime-concept.

src/inbox.rs
Restioson: I agree, I was just mentioning that we must remember to make the switch to a oneshot at some point

src/inbox.rs
thomaseizinger: Potentially although I expect most messages to be composites of basic data types which are all `Unpin.

Restioson: Hm. I feel like it should be possible to remove this bound, though

thomaseizinger: I will try!

src/inbox.rs
thomaseizinger: I recorded it as an issue: https://github.com/Restioson/xtra/issues/168

src/inbox.rs
thomaseizinger: We could not have a trait by duplicating all of the `send_future::Sending` machinery but I figured that is too much code duplication.

src/inbox.rs
thomaseizinger: You were right, we can remove it!

src/inbox.rs
thomaseizinger: I'd prefer to postpone this, esp. because I'd like to discuss https://github.com/Restioson/xtra/issues/94 again.

src/inbox.rs
Restioson: Awesome!

src/inbox.rs
Restioson: Of course :)

src/send_future.rs
```diff
@@ -163,18 +166,20 @@ impl SendFuture<ActorErasedSending, Broadcast> {
...
-    WaitingToSend(Arc<spin::Mutex<WaitingSender<A>>>),
+    WaitingToSend(WaitingSender<M>),
     Done,
 }
 
-impl<A, Rc> Future for Sending<A, Rc>
+impl<A, M, Rc> Future for Sending<A, M, Rc>
 where
     Rc: RefCounter,
+    M: Unpin,
```
Restioson, line 181: Is this bound still required?

thomaseizinger: Yes but in practice, it doesn't matter because `M` will always be an `Arc` or a `Box` here. There is no `Unpin` bound on the `send` and `broadcast` functions on `Address` so users are not restricted on what they can send!

We could probably add a test for that to our new `public_api` test :)

thomaseizinger: We can remove this bound if we remove the `TrySend` trait and duplicate the `Future` impl here. Would you prefer that?

thomaseizinger: Checkout [`e9eb9d3` (#166)](https://github.com/Restioson/xtra/pull/166/commits/e9eb9d3a32b06195ef0061997b77dd69fcd012ce) and let me know what you prefer :)

thomaseizinger: The duplication is actually not as bad as I expected. I think I prefer the version without `TrySend`.

thomaseizinger:
Friendly ping here too @Restioson, I'd appreciate another review :)

Restioson (approved):
Thanks! Feel free to merge when ready

PR #174, opened 2022-08-17
Enforce grouping of imports per module

thomaseizinger:
Most of our codebase seems to be formatted this way, we can enforce
this with a rustfmt setting.


Restioson:
Thanks!

PR #175, opened 2022-08-17
Fix `scoped_actor_task` example

thomaseizinger:
Because of an unknown feature, this example was not compiled as
part of our CI run and we missed it while moving to the new spawn
API.


Restioson:
Good catch!

PR #176, opened 2022-08-17
Use new `xtra::Actor` custom-derive in examples and tests

thomaseizinger:
Open questions:

~- Should we add a dedicated example for how to manually implement the trait?~


Restioson:
> Should we add a dedicated example for how to manually implement the trait?

I think this could be nice to show off using different `Stop` types. E.g, a `stopped()` could return some statistical report about messages handled and errors encountered, or something simpler for the example.

thomaseizinger:
We should merge https://github.com/Restioson/xtra/pull/180 first so that this PR remains a pure refactoring.

Restioson (approved):
Approved in principle - I agree with merging the other PR first though

PR #180, opened 2022-08-30
Rework tests around drop behaviour of actors

thomaseizinger:
Previously, xtra's Actor had different behaviour depending on whether the last `Address` was dropped, `Context::stop` was called or `Context::stop_self` was called.

With the removal of `Actor::stopping` and the simplification of the event loop, these behaviours have been unified. Prior to this patch, the tests written for these cases were a bit outdated.

We simplify these tests and split them into individual ones.

### DIFF
--- a/tests/basic.rs
+++ b/tests/basic.rs
@@ -161,7 +161,7 @@ async fn actor_stops_on_last_drop_of_address_even_if_not_yet_running() {
 }
 
 #[tokio::test]
-async fn actor_stops_on_stop_message_even_if_send_before_started() {
+async fn actor_stops_on_stop_message_even_if_sent_before_started() {
     let (addr, context) = Context::new(None);
     let weak = addr.downgrade();
     let join = weak.join();
--- a/tests/basic.rs
+++ b/tests/basic.rs
@@ -1,7 +1,5 @@
 use std::cmp::Ordering as CmpOrdering;
 use std::ops::ControlFlow;
-use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::Arc;
 use std::task::Poll;
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


tests/basic.rs
```diff
@@ -114,68 +104,71 @@ impl Handler<StopAll> for DropTester {
...
+    tokio::spawn(context.run(StopTester)).await.unwrap();
+
     assert!(!weak.is_connected());
     assert!(join.now_or_never().is_some());
+}
 
-    // Send a stop message before future has even begun
-    let drop_count = Arc::new(AtomicUsize::new(0));
+#[tokio::test]
+async fn actor_stops_on_stop_message_even_if_not_yet_running() {
```
Restioson, line 164: Maybe "even if sent before started" is a better phrasing here?

Restioson (approved):
Just one comment, otherwise LGTM!

tests/basic.rs
thomaseizinger:
```suggestion
async fn actor_stops_on_stop_message_even_if_send_before_started() {
```

tests/basic.rs
```diff
@@ -114,68 +104,71 @@ impl Handler<StopAll> for DropTester {
...
+    tokio::spawn(context.run(StopTester)).await.unwrap();
+
     assert!(!weak.is_connected());
     assert!(join.now_or_never().is_some());
+}
 
-    // Send a stop message before future has even begun
-    let drop_count = Arc::new(AtomicUsize::new(0));
+#[tokio::test]
+async fn actor_stops_on_stop_message_even_if_send_before_started() {
```
thomaseizinger, line 164:
```suggestion
async fn actor_stops_on_stop_message_even_if_sent_before_started() {
```

Issue #116, opened 2022-07-11
Benchmark against mpsc channels


thomaseizinger:
xtra is by design small and thus, I think a big contestor for people making a decision on what to use might be tokio's or futures' mpsc channels, spawned as a task with an endless loop.

It would be interesting to benchmark xtra against such a naive actor implementation.

Restioson:
Example implementation: https://ryhl.io/blog/actors-with-tokio/

Restioson:
> Any thoughts on using an approach like [enum_dispatch](https://crates.io/crates/enum_dispatch) to eliminate the dynamic dispatch overhead?

Thoughts? Yes, many :smile:  plans, not currently. In the xtra-community matrix chat a while ago I sent a gist which I can try to find, which would basically be 'zero dynamic dispatch' xtra (iirc). or maybe it had one box in there somewhere, I don't remember

> Would be able to use off the shelf mpsc channels instead, while keeping the same ergonomics

The reason we have custom channels is actually nothing to do with the message enveloping. For a long time, xtra did indeed use just flume or another off the shelf channel. The issue with the channels is for the multi-actor stuff (message stealing) & broadcast. It's a bit more space and time efficient (and just nicer with `select`ing etc) to roll it all together

Restioson:
Here's the gist i mentioned: https://gist.github.com/Restioson/2b93c1b02ded2aa4422596b3fdd72ef6

Restioson:
> I believe the "static dispatch" you were seeking in your gist would come for free with this enum approach, as the compiler knows to optimize branching away when an enum has 1 variant.

The static dispatch I was seeking was not having to box the message into an envelope which calls the actor handler, so I think this would also cover that if done right

PR #181, opened 2022-08-30
Add benchmark for xtra vs tokio mpsc channels

thomaseizinger:
A first implementation of a benchmark. Here are some first results:

```
increment/xtra/100      time:   [940.73 us 969.06 us 998.71 us]
increment/xtra/1000     time:   [9.6542 ms 9.9258 ms 10.205 ms]
increment/xtra/10000    time:   [97.921 ms 100.49 ms 103.09 ms]
increment/mpsc/100      time:   [14.864 us 14.972 us 15.076 us]
increment/mpsc/1000     time:   [144.32 us 145.35 us 146.34 us]
increment/mpsc/10000    time:   [1.3135 ms 1.3333 ms 1.3532 ms]
```

Fixes https://github.com/Restioson/xtra/issues/116.

### DIFF
--- a/benches/xtra_vs_tokio_mpsc.rs
+++ b/benches/xtra_vs_tokio_mpsc.rs
@@ -69,7 +69,7 @@ fn xtra_throughput(c: &mut Criterion) {
                     },
                     |xtra_address| async move {
                         for _ in 0..num_messages - 1 {
-                            let _ = xtra_address.send(Increment {}).await.unwrap();
+                            xtra_address.send(Increment {}).await.unwrap();
                         }
 
                         xtra_address.send(Stop).await.unwrap();


thomaseizinger:
I am guessing the major speed difference comes from the allocations. `xtra` performs at least two allocations as part of wrapping messages in its envelope, plus there might be allocations within the channel to store waiting senders / receivers.

thomaseizinger:
> True - this is simply the cost you pay for dynamic message dispatch. I would wonder how it'd look if the Tokio example also dynamically dispatched message handlers

Would that be a fair comparison though? The actor as shown in the benchmark is how I'd implement it. With an enum to handle multiple cases.

Restioson:
> Would that be a fair comparison though? The actor as shown in the benchmark is how I'd implement it. With an enum to handle multiple cases.

It's comparing different things - in the dynamic dispatch case what we are really measuring is the channel performance of xtra's internal channel. Previously this could be extrapolated from benches of flume, but this is no longer the case

thomaseizinger:
> > Would that be a fair comparison though? The actor as shown in the benchmark is how I'd implement it. With an enum to handle multiple cases.
> 
> It's comparing different things - in the dynamic dispatch case what we are really measuring is the channel performance of xtra's internal channel. Previously this could be extrapolated from benches of flume, but this is no longer the case

As a user though, I don't really care what `xtra` is using internally? I think `xtra` is quite appealing from an ergonomics perspective. The only thing that these benchmarks should show IMO is that they will not pay a massive price for these ergonomics. 10us per message is still pretty good and likely negligible compared to what the application will actually do.

At least in these benchmarks, the mpsc channel is ~70x faster but I think we are operating at a latency scale here where if that difference matters to you, you are probably going to write everything yourself anyway? I've never worked on applications that were performance critical on that scale so I don't actually know.

Restioson:
> As a user though, I don't really care what xtra is using internally? I think xtra is quite appealing from an ergonomics perspective. The only thing that these benchmarks should show IMO is that they will not pay a massive price for these ergonomics. 10us per message is still pretty good and likely negligible compared to what the application will actually do.

From a user perspective, it is not a useful measurement, I agree. However, it would still be useful to look at for us (the implementors) to see how fast xtra's channel is in comparison to tokio

PR #182, opened 2022-09-01
Fully move from `inbox` to `Chan` terminology

thomaseizinger:
With the introduction of a `Mailbox`, having an `inbox` is confusing when this module actually houses xtra's channel implementation.


src/chan/ptr.rs
```diff
@@ -4,23 +4,23 @@ use std::sync::{atomic, Arc};
...
+use crate::chan::Chan;
 
 /// A reference-counted pointer to the channel that is generic over its reference counting policy.
 ///
 /// Apart from [`TxEither`], all reference-counting policies are zero-sized types and the actual channel
 /// is stored in an `Arc`, meaning this pointer type is exactly as wide as an `Arc`, i.e. 8 bytes.
 ///
 /// This is possible because the actual reference counting is done within [`Chan`].
-pub struct ChanPtr<A, Rc>
+pub struct Ptr<A, Rc>
```
Restioson, line 15: This might be a bit vague?

Restioson (approved):
Just the one comment, otherwise this is a nice rename I think!

src/chan/ptr.rs
thomaseizinger: We can keep it at `ChanPtr`, just need to move away from importing it via the `chan` module then :)

It is something I am not too sure about myself. Generally, I try to not repeat module names within types but that does lead to fairly generic names _within_ the module they are defined in.

thomaseizinger: Is your +1 a "we can merge as is" or do you want me to rename it back to `ChanPtr`?

Restioson: I don't really mind - can go ahead with merge

PR #185, opened 2022-09-08
Remove rx/tx terminology from Debug output

thomaseizinger:
This patch proposes a slightly more user-friendly debug output of addresses and message channels. Instead of rx/tx terminology, we use addresses and mailboxes.


Restioson (approved):

Restioson:
Thanks!

PR #191, opened 2022-09-15
Simplify feature dependencies

thomaseizinger:
1. We only use `futures-util`'s `alloc` feature in the tests.
2. `futures-core` only has one feature and that is alloc and it is on by default.


Cargo.toml
```diff
@@ -15,9 +15,9 @@ rust-version = "1.60.0"
 [dependencies]
 async-trait = "0.1.56"
 catty = "0.1.5"
-futures-core = { version = "0.3.21", default-features = false, features = ["alloc"] }
+futures-core = "0.3.21"
```
Restioson, line 18: Does futures core have no default features?

thomaseizinger: It does: `alloc` :)

Restioson:
While this is a nice simplification, I am not sure if it is necessarily a good idea. It is a little less clear exactly what features are enabled and which aren't, as we are relying on what the default features are, as opposed to writing them out ourself. What do you think?

thomaseizinger:
> While this is a nice simplification, I am not sure if it is necessarily a good idea. It is a little less clear exactly what features are enabled and which aren't, as we are relying on what the default features are, as opposed to writing them out ourself. What do you think?

I would expect `futures-core` to always be extremely lightweight so the risk of having more code "slip in" is quite low I believe. Plus, I'd expect any changes to features to result in a minor version bump as they can be breaking.

Happy to change it back but I also like the simplicity of not needing to set features. I do see though how it can trigger the thought  of "did we vet the default features of this lib?".

I wish default features would have never been introduced :(

Restioson:
> I would expect futures-core to always be extremely lightweight so the risk of having more code "slip in" is quite low I believe. Plus, I'd expect any changes to features to result in a minor version bump as they can be breaking.

Fair enough.

> Happy to change it back but I also like the simplicity of not needing to set features. I do see though how it can trigger the thought of "did we vet the default features of this lib?".

Yea, this was my main point

thomaseizinger:
> > Happy to change it back but I also like the simplicity of not needing to set features. I do see though how it can trigger the thought of "did we vet the default features of this lib?".
> 
> Yea, this was my main point

Do you want me to add a comment?

Restioson:
I think that's a good solution, thanks :)

Restioson (approved):

thomaseizinger:
Resolved! Sorry for the delay.

Issue #188, opened 2022-09-14
`as_either` for `MessageChannel`?


bobdebuildr:
I would like to store various `MessageChannel<M, R, Rc>` in a `Vec<MessageChannel<M, R, Either>>`, where `Rc` could vary. It would be nice to have an `as_either()` function on `MessageChannel` similar to the one for `Address`.

Is there some deeper reason why this isn't possible or shouldn't be done?

bobdebuildr:
I managed to add this function- if you're interested I can open a pull request.

Restioson:
Good catch! This was probably missed in #102. Feel free to open a PR, I'd appreciate that :)

thomaseizinger:
Resolved in #192.

PR #192, opened 2022-09-15
Add `MessageChannel::as_either()`

bobdebuildr:
implements #188

### DIFF
--- a/src/message_channel.rs
+++ b/src/message_channel.rs
@@ -248,7 +248,7 @@ where
 }
 
 /// Functions which apply to any kind of [`MessageChannel`], be they strong or weak.
-impl<M, R> MessageChannel<M, R>
+impl<M, R, Rc> MessageChannel<M, R, Rc>
 where
     M: Send + 'static,
     R: Send + 'static,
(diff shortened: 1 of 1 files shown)


src/message_channel.rs
```diff
@@ -247,6 +247,20 @@ where
     }
 }
 
+/// Functions which apply to any kind of [`MessageChannel`], be they strong or weak.
```
thomaseizinger, line 250: For this comment to be true, you'd need to move away from the defaulted `Rc` variable and actually make this `impl` block generic over all `Rc`s :)

At the moment, it is only available for `Strong`.

thomaseizinger (changes requested):
Thanks! One comment :)

src/message_channel.rs
bobdebuildr: Ah yes, thanks for catching that! I updated the `impl` block accordingly.

Restioson (approved):
Thanks!

thomaseizinger (approved):
Nice work, thanks!

Issue #189, opened 2022-09-15
See if we can simplify `TickFuture`


thomaseizinger:
Follow-up from #133.

Currently, `xtra::tick` does two things:

1. Fetch the next message from the mailbox
2. Pass the message to the actor

(1) is already possible outside of `xtra::tick` by calling `Mailbox::next`. This is the first code-smell.
(2) is a bit more tricky because the `Message` returned by `Mailbox::next` is an opaque struct for users, they can't do anything with it other than instantiate a `TickFuture`.

Part of the reason for that is because `MessageEnvelope` and `BroadcastEnvelope` return a tuple of the span + the handler future and `TickFuture` makes use of that by allowing users to attach to that span.

There may be a simpler / more composable way of expressing this by introducing some API on `Message` that makes it obsolete for users to "call back" into `xtra` via `xtra::tick`.

Restioson:
Is it not the case that if the user wants to poll another future during the loop of an actor, a single, combined receive-and-tick would prevent this?

thomaseizinger:
> Is it not the case that if the user wants to poll another future during the loop of an actor, a single, combined receive-and-tick would prevent this?

I am not advocating for it to be combined. In fact I want to it to be two seperate actions but they should compose well together without the need for `xtra::tick`.

Restioson:
Could you give an example on how an API for this could look?

thomaseizinger:
I put up a PR here: #194.

PR #194, opened 2022-10-28
Remove `xtra::tick` in favor of `Message::dispatch_to`

thomaseizinger:
Resolves https://github.com/Restioson/xtra/issues/189.


thomaseizinger:
If we end up merging this, then I think the public API of `0.6` is _really_ good.

xtra/src/chan/waiting_sender.rs
```diff
@@ -7,6 +7,7 @@ use std::task::{Context, Poll, Waker};
 use crate::chan::{HasPriority, Priority};
 use crate::Error;
 
+#[must_use = "Futures do nothing unless polled"]
 pub struct WaitingSender<M>(Arc<spin::Mutex<Inner<M>>>);
```
Restioson, line 11: Nice catch!

Restioson (approved):
Thanks!

PR #207, opened 2023-09-26
Remove ordered mailbox and rename priority to unicast

### LINKED ISSUE #94: Re-consider separate, ordered mailbox

thomaseizinger:
This aims to resolve #94 by renaming `split_receiver` to `detach`. This wording has prior art in the ecosystem, for example:

- https://docs.rs/smol/1.3.0/smol/struct.Task.html#method.detach
- https://docs.rs/tokio/1.32.0/tokio/task/join_set/struct.JoinSet.html#method.detach_all

Both of these are about "running a task in the background".

As a 2nd step and with the terminology resolved, we remove the ordered mailbox and rename the priority one to unicast as a counterpart to broadcast.


Restioson:
Happy to merge when passing CI - I assume will need rebase from master.

Thinking about this again with fresh eyes, I think it really does make a lot of sense. I'm not sure if this was said by either of us before, but this kind of mimics  atomics in some way (?). At least, that's how my mental model for this is now.

If you have two threads which are doing atomic operations on the same thing, the order of their atomic operations relative to eachother, if they are both using relaxed orderings, will be unspecified. They will happen, but they could interleave arbitrarily. The compiler is _also_ free to rearrange operations on that thread.

Previously, ordered sends worked the same, except that operations on one thread would still be ordered. This is nice, but as we've discussed can lead to backpressure footguns and is also harder to support. Plus, it's possible to emulate previous behaviour with new behaviour: you could always
```rust
// look ma, just like `detach` but with multiple messages
tokio::spawn(async {
    addr.send(1).await;
    addr.send(2).await;
});
```

thomaseizinger:
> Thinking about this again with fresh eyes, I think it really does make a lot of sense. I'm not sure if this was said by either of us before, but this kind of mimics atomics in some way (?). At least, that's how my mental model for this is now.

I am glad that we arrived at this! I agree with the atomics view. The similarity is not surprising when you think about it: `Address` is a way of writing to an actor without requiring `&mut self`.

thomaseizinger:
@Restioson I had to adapt a few tests, please have a look1

xtra/tests/basic.rs
```diff
@@ -6,6 +6,7 @@ use std::time::Duration;
 use futures_util::task::noop_waker_ref;
 use futures_util::FutureExt;
 use smol_timeout::TimeoutExt;
+use tokio::task::JoinSet;
```
Restioson, line 9: side note: it's funny how many async crates we're mixing and matching here, haha

Restioson (approved):

PR #210, opened 2023-09-29
Update lockfile before computing cache key

thomaseizinger:
`Swatinem/rust-cache` considers the lockfile it finds in the repo for its cache key. If we update it _after_ restoring the cache, the cache key will never match exactly.


Restioson (approved):

Restioson:
By the way, what else do we want to get in before releasing?

thomaseizinger:
> By the way, what else do we want to get in before releasing?

I think we are good! I'll see to tidy up the changelog and README a bit. I also want to write a blogpost about the release :)

Issue #202, opened 2023-02-15
Experiment with nightly async fn in traits


thomaseizinger:
Both `Handler` and `Actor` are currently using `async-trait`. We should investigate how the recent nightly-only features regarding async fn in traits helps us.

PR #211, opened 2023-09-29
Remove `async_trait` macro in favor of AFIT & RPITIT

thomaseizinger:
Resolves: #202.

### DIFF
--- a/xtra/src/lib.rs
+++ b/xtra/src/lib.rs
@@ -53,6 +53,7 @@ pub mod refcount {
 /// The [`Actor`] custom derive takes away some boilerplate for a standard actor:
 ///
 /// ```rust
+/// # #![feature(async_fn_in_trait)]
 /// #[derive(xtra::Actor)]
 /// pub struct MyActor;
 /// #
@@ -65,6 +66,7 @@ pub mod refcount {
 /// This macro will generate the following [`Actor`] implementation:
 ///
 /// ```rust,no_run
+/// # #![feature(async_fn_in_trait)]
 /// # use xtra::prelude::*;
 /// pub struct MyActor;
 ///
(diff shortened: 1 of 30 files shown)


xtra/src/lib.rs
```diff
@@ -114,16 +113,16 @@ use crate::recv_future::Message;
...
     /// Handle a given message, returning its result.
-    ///
-    /// This is an [`async_trait`](https://docs.rs/async-trait).
-    /// See the trait documentation to see an example of how this method can be declared.
-    async fn handle(&mut self, message: M, ctx: &mut Context<Self>) -> Self::Return;
+    fn handle(
+        &mut self,
+        message: M,
+        ctx: &mut Context<Self>,
+    ) -> impl Future<Output = Self::Return> + Send;
```
thomaseizinger, line 125: A limitation of the current feature is that there is no "return-type notation", meaning users of this trait cannot specify, whether or not this future is `Send`. If we would just use `async fn` here in the declaration, this `Future` would not be `Send` and thus we would not be able to use `xtra` on multi-threaded executors.

From reading https://github.com/rust-lang/rust/pull/115822, this is what is most likely going to be stablized. It is still quite cool that this just works now! :)

thomaseizinger: As we can see in the rest of the codebase, implementations can just use `async fn` and they will inherit the `Send` requirement.

Restioson (commented):
Very cool! I'm surprised that MessageEnvelope didn't need any changes

thomaseizinger:
https://github.com/rust-lang/rust/pull/115822 just merged which will be stable in Rust 1.75 on December 28th! :partying_face:

Restioson:
Woo! Awesome!

thomaseizinger:
@Restioson I think this is now ready for being merged once `1.75` is released.

Restioson:
Awesome! :tada:

thomaseizinger:
@Restioson This is now ready to be merged!

Restioson (approved):

thomaseizinger:
Shall we release `0.6` finally? :partying_face:

PR #215, opened 2023-12-07
fix: update doctests to use smol feature

### LINKED ISSUE #214: Deprecated feature in doctests
Looks like there are some broken doctests which contain old feature gates.

Example in `select` doctest

https://github.com/Restioson/xtra/blob/537cfad5115865ff0e510188df8413095eabe0e4/xtra/src/lib.rs#L300-L305

sinui0:
This PR updates the doctests on `select` and `join` functions to get rid of the deprecated `with-smol-1` feature gate and uses the new `smol` feature instead.

Closes #214 

Should `smol` be added into `[dev-dependencies]` so this sort of thing isn't possible? Doctests should ideally run fully without having to enable a feature.

### DIFF
--- a/xtra/src/lib.rs
+++ b/xtra/src/lib.rs
@@ -297,13 +297,12 @@ where
 ///     }
 /// }
 ///
-/// # #[cfg(feature = "with-smol-1")]
-/// # smol::block_on(async {
-/// let addr = MyActor.create(None).spawn(&mut xtra::spawn::Smol::Global);
-/// assert!(addr.is_connected());
-/// assert_eq!(addr.send(Selecting).await, Ok(true)); // Assert that the select did end early
-/// # })
-///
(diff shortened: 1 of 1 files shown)


thomaseizinger:
> Should `smol` be added into `[dev-dependencies]` so this sort of thing isn't possible? Doctests should ideally run fully without having to enable a feature.

We can't do that because we are exposing helper functions here: https://github.com/Restioson/xtra/blob/e96ae544d48b5b0154f3f0871f06a137e09f6005/xtra/src/spawn.rs#L31-L44.

thomaseizinger:
Can you update to latest master? That should fix CI I believe!

Restioson:
Happy to merge after passes CI

thomaseizinger (approved):

PR #216, opened 2023-12-07
chore: add explicit macros feature to manifest

sinui0:
This PR adds an explicit `macros` feature to the manifest.

Typically one will look at the manifest `[features]` section to quickly identify which features are available (especially if not documented in the top-level crate docs). Searching for implicit features via optional dependencies leaves something to be desired.


thomaseizinger (approved):

thomaseizinger:
Thanks and sorry for the delay!

PR #218, opened 2024-02-02
Release 0.6.0

Restioson:


Restioson:
@thomaseizinger: I've released from this branch now - can we merge these in? minimal set of changes to get it to work

macros/Cargo.toml
```diff
@@ -1,7 +1,11 @@
 [package]
 name = "xtra-macros"
-version = "0.1.0"
+version = "0.6.0"
```
thomaseizinger, line 3: Why does this need to track the version of the root crate?

thomaseizinger:
> @thomaseizinger: I've released from this branch now - can we merge these in? minimal set of changes to get it to work

For the future, I'd suggest to not release from branches but only after they are merged. With squash merging, the commits in the branch never land on master, meaning you can't have a tag now that directly corresponds to the released code.

Restioson:
Are you okay to just rebase merge this in to master then? Thanks for the tip, I didn't anticipate that

macros/Cargo.toml
Restioson: Arguably it does not, but I think this is a standard set in other crates - serde and serde_derive seem pegged, as do bevy and bevy_internal/bevy_dylib/bevy_input.

Otherwise, the option is to maintain a table like this;
image

thomaseizinger:
> Are you okay to just rebase merge this in to master then? Thanks for the tip, I didn't anticipate that.

Yes I think this would be a good idea! :)

macros/Cargo.toml
thomaseizinger: I guess it does makes things easier while we are in `0.x` land and after `1.x`, the difference doesn't matter too much :)

thomaseizinger (approved):

PR #221, opened 2024-02-15
chores: Bump smol version to 2.0

### LINKED ISSUE #220: Bump smol version to `2.0`
Current version of xtra locks on `1.1` minimum, which makes cargo compile it twice if you already had smol on your crate.

Altair-Bueno:
This should fix #220. The changes include removing smol-potat in favor of smol::block_on. However, this commit does not remove nor update smol-timeout as it is heavily used on tests and the crate seems to be unmaintained. Nonetheless tests keep working.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -48,29 +48,41 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-channel"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f28243a43d821d11341ab73c80bed182dc015c514b951616cf79bd4af39af0c3"
+dependencies = [
+ "concurrent-queue",
+ "event-listener 5.0.0",
+ "event-listener-strategy 0.5.0",
+ "futures-core",
+ "pin-project-lite 0.2.13",
+]
+
+[[package]]
 name = "async-executor"
 version = "1.5.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2c1da3ae8dabd9c00f453a329dfe1fb28da3c0a72e2478cdcd93171740c20499"
 dependencies = [
- "async-lock",
+ "async-lock 2.8.0",
  "async-task",
  "concurrent-queue",
  "fastrand 2.0.1",
- "futures-lite",
+ "futures-lite 1.13.0",
  "slab",
 ]
 
 [[package]]
 name = "async-fs"
-version = "1.6.0"
+version = "2.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "279cf904654eeebfa37ac9bb1598880884924aab82e290aa65c9e77a0e142e06"
+checksum = "bc19683171f287921f2405677dd2ed2549c3b3bda697a563ebc3a121ace2aba1"
 dependencies = [
- "async-lock",
- "autocfg",
+ "async-lock 3.3.0",
  "blocking",
- "futures-lite",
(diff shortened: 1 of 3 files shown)


Restioson:
Hey, thanks for the PR! I wonder if we would want to keep smol 1.0 support under a different feature (e.g. smol-1)?

thomaseizinger:
> Hey, thanks for the PR! Really sorry for getting back to this so late, studies have been all-consuming this year. I wonder if we would want to keep smol 1.0 support under a different feature (e.g. smol-1)?

As long as there is a release with smol-1 support, I wouldn't bother too much :)

Restioson:
That's fair, since I suppose we haven't really made any other changes

xtra/examples/message_stealing_multiple_actors.rs
```diff
@@ -54,3 +54,7 @@ async fn main() {
     // Give a second for everything to shut down
     std::thread::sleep(Duration::from_secs(1));
 }
+
+fn main() {
+    smol::block_on(run())
+}
```
Restioson, line 60: Would be happy to merge if we can add back the newline :)

Altair-Bueno:
So, almost a year later, I noticed this PR was still open and that I missed your review. Sorry for that. I ran `cargo fmt --all` so that the style is ok.

Restioson (approved):

Restioson:
Apologies - this dropped off my radar too!

Restioson:
Hmm, looks like a build failure - I think smol relies on async_task?

Altair-Bueno:
I think it had something to do with Cargo.lock being modified in both branches so it was stuck in an invalid state. I recreated the lock file with the latest dependencies. 

I would advise removing the lockfile, as it mostly does nothing for libraries, and switch to something like https://crates.io/crates/cargo-minimal-versions to validate the minimum requirements.

thomaseizinger:
> I would advise removing the lockfile, as it mostly does nothing for libraries, and switch to something like https://crates.io/crates/cargo-minimal-versions to validate the minimum requirements.

Having the lockfile ensures deterministic CI runs, which is important in the presence of e.g. semver-incompatible bumps etc

It is also the official recommendation from the Rust folks: https://blog.rust-lang.org/2023/08/29/committing-lockfiles/

Cargo.lock
```diff
@@ -130,48 +141,48 @@ dependencies = [
...
  "signal-hook-registry",
  "slab",
- "windows-sys",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
 name = "async-std"
-version = "1.12.0"
+version = "1.13.1"
```
thomaseizinger: Why is async-std getting bumped when we are only changing smol? 🤔

thomaseizinger:
The diff on the lockfile seems a bit excessive to me. Can you checkout the one from main and redo the bump to make that smaller? Thanks!

Altair-Bueno:
> It is also the official recommendation from the Rust folks

Oh, I guess I was using the old guidelines. Thanks for sharing.

> The diff on the lockfile seems a bit excessive to me. Can you checkout the one from main and redo the bump to make that smaller? Thanks!

Done. I had to use Rust 1.75 as 1.88 was failing for wasm_bindgen with the following error:

```
error: older versions of the `wasm-bindgen` crate are incompatible with current versions of Rust; please update to `wasm-bindgen` v0.2.88
```

Restioson:
LGTM, happy for you to merge @thomaseizinger when you're happy

thomaseizinger:
Is the CI error reproducable locally?

PR #226, opened 2024-07-03
feat: implement Hash for MessageChannel

jcdickinson:
This adds the `Hash` trait to `MessageChannel`. I need this in my project to use a `MessageChannel` as a key in a `HashMap` (as actor types are generally unknown for my use case). Pairs with the existing `PartialEq`.

### DIFF
--- a/xtra/src/message_channel.rs
+++ b/xtra/src/message_channel.rs
@@ -3,6 +3,7 @@
 //! the message type rather than the actor type.
 
 use std::fmt;
+use std::hash::{Hash, Hasher};
 
 use crate::address::{ActorJoinHandle, Address};
 use crate::chan::RefCounter;
@@ -207,6 +208,17 @@ where
     }
 }
 
+impl<M, R, Rc: RefCounter> Hash for MessageChannel<M, R, Rc>
+where
+    M: Send + 'static,
+    R: Send + 'static,
+    Rc: Send + 'static,
+{
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.inner.hash(state)
+    }
+}
+
 impl<M, R, Rc> Clone for MessageChannel<M, R, Rc>
 where
     R: Send + 'static,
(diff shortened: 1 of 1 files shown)


thomaseizinger (commented):
Concept ACK but can you add unit tests please?

jcdickinson:
Updated. The failing check seems unrelated to my changes, and I can't repro it locally.

xtra/src/message_channel.rs
```diff
@@ -366,4 +380,64 @@ where
...
+    }
+
+    impl Handler<TestMessage> for TestActor {
+        type Return = ();
+
+        async fn handle(&mut self, _: TestMessage, _: &mut crate::Context<Self>) -> Self::Return {}
+    }
+
+    #[test]
+    fn hashcode_and_partial_eq() {
```
thomaseizinger, line 419: This only tests hash-code, right?

Would you mind adding one for `PartialEq` too please? 🙏 (I do realise that we should have had one for that from the beginning.)

thomaseizinger:
> Updated. The failing check seems unrelated to my changes, and I can't repro it locally.

I'll see if I can quickly fix that.

xtra/src/message_channel.rs
jcdickinson: Hashcodes are permitted to collide (they only guarantee inequality), so HashSet will call PartialEq (via Eq) during the second call to insert.

jcdickinson: Hmm, I should also check the negative case, making the change now.

jcdickinson: I have completely changed the approach and avoid using `HashSet` as a property test. We know that the source data for the hashcode should be different, so that is tested directly.

thomaseizinger (approved):
Awesome stuff, thank you for adding these tests!

thomaseizinger:
I've opened https://github.com/Restioson/xtra/pull/227 to fix CI need to wait for @Restioson to approve it!

Restioson (approved):
Apologies for getting to this late; I've been at a conference

thomaseizinger:
@jcdickinson Can you rebase on latest master?

jcdickinson:
Done!

